PR #418, opened 2015-08-17
chore(.travis.yml): include gitter webhook

gdi2290:


michaelbpaulson:
@gdi2290 We deleted the gitter chat as we are now open to the world.  We would prefer issues and stack overflow questions.

gdi2290:
alright that makes sense

PR #431, opened 2015-08-17
Run travis builds as a docker container

megawac:
Also, should the travis build be moved over from magnum to the docker version for open source projects?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ branches:
     only:
         - master
 
+# Run as a docker container
+sudo: false
+
 # Update npm to the latest version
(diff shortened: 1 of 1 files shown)


gdi2290:
:+1: this speeds up build significantly

jhusain:
@sdesai Does this work for you?

sdesai:
Yep. It's fine. It's running in a docker container even without this change (e.g: https://travis-ci.org/Netflix/falcor/jobs/76140461), but it's fine to merge in.

Doesn't hurt to force it to "sudo:false" which will highlight when/if we do need sudo explicitly for something in the build.

PR #450, opened 2015-08-18
bumping version number

michaelbpaulson:


michaelbpaulson:
For the sake of getting rid of the checksum error.

PR #458, opened 2015-08-19
Remove webpack notice

adi-ads:
RxJS issue has been resolved and released, tested with falcor and works well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,23 +8,6 @@
 
 **This release is a developer preview.** We are looking for community help to track down and fix bugs. We are also looking for help integrating with existing MVC frameworks, as well as ports to other platforms.
 
-## Important Note for Webpack Users
-
-If you're including falcor in your app, via npm and `require('falcor')`, and you're building a browser bundle for your app with Webpack, you'll need to add an alias entry for the 'rx' module in your webpack config, to avoid this RxJS bug: 'https://github.com/Reactive-Extensions/RxJS/issues/832'. You may already have such an entry, if you're using RxJS already. An example is below:
-
-In webpack.config.js:
-
-```js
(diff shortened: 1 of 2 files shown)


Dashed:
If it works, shouldn't this be removed then? https://github.com/Netflix/falcor/blob/master/webpack.config.js

adi-ads:
Good point, deleted.

sdesai:
Did you have a chance to test this off of npm?

When I looked at it last night, it was fixed on RxJS#master, and I tested it from there (https://github.com/Reactive-Extensions/RxJS/issues/832#issuecomment-132407081), but wanted to make sure it's working off of npm, out of the box, before removing the notice and config.

sdesai:
It's also worth noting that this will bump up the RxJS dep from 2.x to 3.x, so that may need some validation, outside of the webpack issue [ unit tests, performance, regressions etc. ]

adi-ads:
Only did it with local git clones, will try out with npm

sdesai:
Cool. Thx. Although as mentioned above we may need to gate removing the webpack notice/config on determining whether or not we want to upgrade to RxJS 3.x in general, which is a broader decision, outside of the WebPack issue (for example https://github.com/Netflix/falcor/issues/464)

dallonf:
This fix _is_ still needed for the version of RxJS listed in package.json, FYI, otherwise you'll run into bizarre issues where Observables don't have `reduce()` functions and the like.

I didn't RTFM. Don't be like me.

sdesai:
I'm closing this out. We'll need the notice in place until we complete removal of RX as a core falcor dependency. See: https://github.com/Netflix/falcor/issues/465

PR #461, opened 2015-08-19
Reuse slice

megawac:


michaelbpaulson:
I hate to do this but a lot of the code you help made smaller has been removed. You should check out `walkPath.js` it's been reduced a lot since path maps are no longer valid input.

PR #466, opened 2015-08-21
export browser.js for browserify to require

jcouyang:
When I require falcor and bundle with browserify, it dosen't require
in HttpDataSource, which I really want to bundle in my app.

the idea is by export the `browser.js` file so basically I can `require('falcor/browser')` instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "build",
     "dist",
     "lib",
+    "browser.js",
     "test"
   ],
   "keywords": [


gdi2290:
good idea, you can also import https://www.npmjs.com/package/falcor-http-datasource that lives in  https://github.com/Netflix/falcor-http-datasource which also has node support

jhusain:
Curious why we want two ways to do this? Should requiring the package be the only way we should support? @gdi2290 what's the rationale here?

gdi2290:
If people still prefer one require statement (only require `falcor`) but it's all preference 
```es6
var falcor = require("./lib");
var jsong = require("falcor-json-graph");

falcor.atom = jsong.atom;
falcor.ref = jsong.ref;
falcor.error = jsong.error;
falcor.pathValue = jsong.pathValue;

falcor.HttpDataSource = require("falcor-http-datasource");

module.exports = falcor;
```
It's probably better in the long run for people to use the modules `falcor-json-graph` and `falcor-http-datasource` separately if they want to bundle the repo on their own. Doing so will probably encourage people to create their own DataSource or json-graph utils

jcouyang:
yeah, agree with `require('falcor')` for only falcor, but datasource or json graph is just like addons to me. for example we can require only react by `require('react')`, alternatively we should able to require all addons together with `require('react/addons')` as well

sfrdmn:
I think it's reasonable for developers using browserify to expect the same package to come out of `require('falcor')` as they would get using `<script src="falcor.browser.js"></script>`

All that would need to be done would be to add a `browser` field in `package.json` pointing to whichever entry point was used to build `falcor.browser.js` [link](https://github.com/substack/node-browserify#browser-field)

michaelbpaulson:
I think that this is reasonable.  Sorry, I never noticed this for some odd reason.

michaelbpaulson:
One problem @jcouyang .  Master wont be released for a couple weeks (working towards 1.x release).  With that in mind, you should probably create this PR against branch 0.x _as well (keep this one up)_ (as 0.x is not mergeable with master currently).

sfrdmn:
Nitpick, but I think it's a bit more idiomatic in the browserify ecosystem to actually use the `browser` field, rather than publish an additional file. Well, the difference is just `require('falcor/browser')` vs `require('falcor')`

michaelbpaulson:
@sfrdmn The major problem I have with that is there are inevitably be users who are going to include falcor-http-datasource when they do not want to.  Its better in my opinion to opt into more than to opt in to less. As far as idiomatic goes, that is something I am unsure of, as its really hard to measure.

PR #469, opened 2015-08-21
Update ESLint to latest

carlosperez:
We noticed in bitHound that the version of ESLint Falcor is using is outdated. Because updating to the latest version of ESLint also requires a config change, we updated that too.

- [x] Update `gulp-eslint` npm package to latest 
- [x] Update `.eslintrc` to match the new config format 
- [x] Update 2 deprecated rules to the new rule names

See justification for these changes here https://www.bithound.io/github/bithound/falcor/eslint.update


baer:
A few comments here:

1. ESLint 1.x ignores deprecated rules so you really don't need to do that anymore.
2. NPM 3 is right around the corner and will require that app/lib/package developers resolve peerDependencies on manually. You might as well pull in ESLint itself since you're going to have to in a few more weeks anyway to support Node 4 (the io+node reconciliation branch) which will ship with NPM 3.
3. It might be worth using a more shared config so you don't have to think about these things. I'm partial to [eslint-config-defaults](https://github.com/walmartlabs/eslint-config-defaults) since I maintain it but there are plenty of good ones.

PR #474, opened 2015-08-23
Fix the falcor-hapi link

carsonmcdonald:
Looks like the project moved.


dallonf:
I wonder if it's still a "third-party integration" if it's on the Netflix org?

Issue #465, opened 2015-08-20
Remove Rx from core


jhusain:
We plan to evaluate removing RX from the core of Falcor. There are several things that motivate this decision:

1. speed
2. debuggability
3. file size

We will be doing some performance analysis over the next few weeks.

@yunong

yunong:
Sounds good. I can help with performance analysis as needed.

trxcllnt:
@jhusain @ktrott how'd we decide on this? should I stop work on integrating RxJS-Next?

ewinslow:
+1 on filesize concerns. My eyes went like O_O when I saw how large Falcor is right now.

ktrott:
@trxcllnt No decision has been made yet. You should proceed with your work as we're using that to evaluate the performance of the ReactiveX/RxJS Next project so that work is tied to both projects.

trxcllnt:
The preliminary upgrade to RxJS-Next is done and available in my forks:

https://github.com/trxcllnt/falcor/tree/rx-next
https://github.com/trxcllnt/falcor-router/tree/rx-next

You’ll have to manually build RxJS-Next, since we’re not presently committing the “dist” folder.

For falcor:

```
~ cd ./node_modules/RxJS-Next
~ npm i
```

For falcor-router (have to build both falcor and falcor-router’s RxJS-Next packages):

```
~ cd ./node_modules/RxJS-Next
~ npm i
~ cd ../falcor
~ npm i
~ cd node_modules/RxJS-Next
~ npm i
```

michaelbpaulson:
I will be finishing up the get request cycle to have rx completely removed from it.  This will be second step towards NoRX in falcor.

tivac:
I was curious about size savings, so did a quick test using browserify. `-min.js` files have been minified with `uglify-js`, and `.gz` files were gzipped with `gzipme`.

```
513,045 falcor.js
278,833 falcor.min.js
276,966 falcor-norx.js
161,130 falcor-norx.min.js
 98,226 falcor.js.gz
 57,624 falcor-norx.js.gz
 54,552 falcor.min.js.gz
 32,932 falcor-norx.min.js.gz
```

Obviously this doesn't take into account the size of whatever would replace rx.js, but it is interesting IMO.

michaelbpaulson:
[...]

PR #482, opened 2015-08-25
Get request

michaelbpaulson:
Get request has been stripped down to prove out the no-rx version of falcor.
#465


lib/request/complement.js
```diff
@@ -0,0 +1,30 @@
+var hasIntersection = require('falcor-path-utils').hasIntersection;
+
+/**
+ * creates the complement of the requested and optimized paths
+ * based on the provided tree.
+ *
+ * If there is no complement then the this is just a glorified
```
michaelbpaulson: I bet i could use some sort of flag and `slice` to improve the performance where if there is no complement then its a noop `n` loop.

michaelbpaulson:
@jhusain @ktrott @trxcllnt @sdesai  If you could review some of this. 

I have created the following:
RequestQueueV2
RequestQueueRx

RequestQueueV2 has the get method created in which does batching and deduping (proper subsets (same behavior as before)) using GetRequestV2

RequestQueueRx takes the functionality from RequestQueue's `set` and `call` and RequestQueueV2's get implementation, wraps it in Rx, and creates an identical interface to the original request queue with the no-rx approach.

I wrote a lot of unit tests to ensure this behavior, thats the bulk of the additional code.

lib/request/complement.js
michaelbpaulson: Fixed this,  Its now no-copy on complete complement (no intersection).

michaelbpaulson:
Results: The request performance tests [found here](https://github.com/michaelbpaulson/falcor/blob/get-request/performance/tests/get/request-queue.js#L51) show about a 1.75 - 4x depending on device / browser.  

Bottlenecks going forward in order
1. merge output
2. get rx
3. get core

michaelbpaulson:
@all there is one issue with this request, we temporarily increase minified size (by a few k).  But this gets us 1 step (biggest step) closer to removing rx dependency.

PR #485, opened 2015-08-26
The cache generator and tests that use it.

michaelbpaulson:
This is for performance testing.


michaelbpaulson:
This is perf only stuff

PR #486, opened 2015-08-26
removed the dup rx deps.

michaelbpaulson:
This reduces the min file size.


michaelbpaulson:
The get-request accidentally added rx 2x to the min build with
`require('rx')` and `require('rx/dist/rx')`

Issue #453, opened 2015-08-18
Code clean-up: Remove selector function / output PathValues code throughout code base


ktrott:
Selector functions were removed before the open source from the public API. We need to clean up and remove it from the code base.

michaelbpaulson:
I am going through `get` right now.  I have hopefully done as little refactoring as my emotions will allow me :)

michaelbpaulson:
@jhusain Are we interested in the following.

?
get(Array|String)
getWithPaths(Array)
getWithJSON(PathMap) 

It would make this clean up worth it.

jhusain:
Please don't change any public APIs yet. I would like to see performance numbers that show that overloading is expensive. 

You can get rid of toPathValues as far as I'm concerned.

trxcllnt:
@jhusain don't we need that for streaming values from a dataSource?

jhusain:
We can stream JSON. In practice values rarely arrive individually from the server, they arrive in bursts. Under the circumstances streaming JSON should be adequate.

PR #489, opened 2015-08-28
Removal of selector and pathValues with slight get clean-up.

michaelbpaulson:
This removes the selector and path values.  Most of the edits were to the unit tests.


michaelbpaulson:
#368 #453

PR #505, opened 2015-08-31
Primed perf

michaelbpaulson:
There are a few linting issues that I fixed as well.


michaelbpaulson:
Results show minimal performance loss for out of cache get requests (not on device though) and about 50% speed up on in cache get requests both on device and in node.

Issue #506, opened 2015-08-31
Remove RX from get request


michaelbpaulson:
The get request currently uses Rx for its async management.  Instead a more performant option is required.  Remove RX from get and instead implement a disposable-callback version that is called behind a ModelResponse and is disposable through the subscribe of the ModelResponse.

michaelbpaulson:
#465

PR #508, opened 2015-08-31
Norx get

michaelbpaulson:
The no Rx get.


michaelbpaulson:
#506 #465

michaelbpaulson:
Here are some numbers for a row get (lolomo, 0, {to: 9}, item, summary)
Master with primed cache: 1
Master with no cache: 1
NoRX with primed cache: 1.52
NoRX with no cache: 1.2

It appears that set is probably one of the bigger slow downs in the no cache scenario.

michaelbpaulson:
@steveorsomethin It turns out its about 20 - 40% faster with out of cache requests and 50-70% faster in cache requests compared to our **last release**.  I think we can squeeze and extra 5 - 10% out with simple API change.  I am going to do some extra unit testing around get to ensure that we have no holes.

michaelbpaulson:
@jhusain @ktrott @sdesai I am going to merge this unless there are any objections.

sdesai:
I noticed the latest set of get changes (in master) seem to be throwing, handling the get request for the browser build (falcor-http-datasource/onXhrLoad). Was just about to file an issue.

I think we should resolve that, before merging in more changes around Get (unless it's fixed in this branch/PR, by virtue of the rx removal/refactor).

Looks like somewhere along the chain onNext, onError, onComplete functions aren't being normalized to an observer object, which is what onXhrLoad is expecting. The "Getting Started" snippet, falcor-express-demo shows this bug.

michaelbpaulson:
@sdesai sounds like a great integration test.  Ill see what I can find.

michaelbpaulson:
@sdesai could you file a ticket with step by step instructions for repro.  If I get a change I'll see if its still broken.

sdesai:
Will do. It's basically the "Getting Started" section, with a local dist build (as opposed to the hosted version off github)

michaelbpaulson:
This will take some time to make this work again due to a small issue, I'll take care of that some time into the future

michaelbpaulson:
@sdesai I have found the root cause of #512 and have a PR to add a unit test to master #515.  Is this safe to merge?  If so, please go for it, else once that problem is fixed could you please merge this?  Thanks.

sdesai:
@jhusain - I'm fine merging this in now that we've addressed #512 - my concern was introducing more flux with the basic integration test being broken.

I don't have enough context yet to review the implementation. Can you take a look at it and we'll merge it in if it looks good.

Issue #472, opened 2015-08-22
Code cleanup: Refactor set, move graph operations to falcor-json-graph


trxcllnt:
@ktrott @jhusain @michaelbpaulson @sdesai @msweeney 

Based on what we've learned in the last three months, the internal set operations are no longer required to build every type of output format, and they don't need to build missing paths. This means most of the internal set implementations are obsolete (and bloating the library), and even the three that are still used are doing more work than they need to.

We only need three set operations:
1. `setJSONGraph`,
2. `setPathMaps`,
3. `setPathValues`

I've built the three new set operations and removed the obsolete modules. Some of these modules are duplicated between falcor and falcor-router, and presumably other projects that want to perform json-graph manipulation. I propose factoring the internal json-graph manipulation modules into the [falcor-json-graph](https://github.com/Netflix/falcor-json-graph) project.

Removing the dead code and factoring out duplicated should further significantly reduce the built file size.

michaelbpaulson:
@trxcllnt could you put those functions in https://github.com/Netflix/falcor-path-utils 

That is where all of our path operations are going (collapse is there currently).

michaelbpaulson:
This removes code output, but does not move core code to falcor-json-graph.  That can be discussed at a later date.

PR #511, opened 2015-09-01
Merge minimal set.

trxcllnt:
The new minimal set and invalidate implementations. Addresses #472.


jhusain:
@michaelbpaulson any chance you can take a look at this? If not, I understand :-)

ktrott:
This PR is quite large and the primary person to review is on paternity leave so just want to provide context here that we'll need some time to review this.

michaelbpaulson:
I think this looks good.  There are some serious changes, but the walks look good, and so do the unit tests so I am ok with merging.  Ill merge this now.

PR #515, opened 2015-09-03
Here is the unit test intentionally failing to show the problem is with http-datasource

michaelbpaulson:
Here @sdesai


test/integration/express.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    it('should be able to perform the express demo.', function(done) {
+        var model = new falcor.Model({
+            source: new falcor.HttpDataSource('http://localhost:1337/model.json')
+        });
+
+         model.
+             get('genrelist[0..5].titles[0..5].name').
+             subscribe(function (data) {
```
michaelbpaulson: The subscribe.onNext is never called, but onError is

sdesai:
Closing this out, and replacing it with 516 (which has some minor tweaks to get it in it's final form)

PR #516, opened 2015-09-04
Demo get integration test

sdesai:
Replaces #515 (adds final assertion, and changes describe.only, to describe).

This test passes with PR https://github.com/Netflix/falcor-http-datasource/pull/19 and fails without it.


sdesai:
This is failing because it pulled in 'falcor-router-demo' as a dependency for integration testing, and 'falcor-router-demo' has older 'falcor' and 'falcor-json-graph' versions as dependencies, which still reference git+ssh:// urls [ which travis can't clone ]. Planning to bump up the falcor-router-demo deps, and publish a new version.

PR #518, opened 2015-09-04
Fixes CallResponse to prepend the bound path to the callPath when calling the data source.

trxcllnt:
@jhusain @michaelbpaulson caught this today, does this fix look correct?


jhusain:
looks good to me. @michaelbpaulson?

michaelbpaulson:
+1

PR #522, opened 2015-09-06
Adding all build target so Router can be used in browser

jhusain:
Creating an "all" js which contains the Router.

@sdesai 
@michaelbpaulson: can you take a look at this? Pay special attention to the package.json. Note that I've set both the http-data-source and router dependencies to be anything larger than the current version. I also updated the version to the latest in npm. Is this the right thing? Previously these dependencies were fixed to a version.


build/gulp-build.js
```diff
@@ -28,16 +29,33 @@ gulp.task('dist.browser', ['clean.dist'], function(cb) {
...
+        browserifyOptions: { standalone: 'falcor' }
+    }, cb);
+});
 
 function build(options, cb) {
     options = _.assign({
         file: '',
         browserifyOptions: {},
-        outName: 'falcor.browser',
+        outName: options.outName,
```
michaelbpaulson: technically you could just remove the line since lodash will just assign it.

PR #529, opened 2015-09-08
Walk faster

michaelbpaulson:
The faster walk has one of the biggest effects on the speed of the get algorithm.  A simple removal of function calls when not needed has made a huge difference.


michaelbpaulson:
Roughly a 20 - 33% speed up on primed cache requests depending on the platform (node vs device).

PR #468, opened 2015-08-21
Update outdated npm packages to latest

carlosperez:
Since bitHound started scoring dependencies, we noticed a dramatic drop in score for the Falcor project. This was caused by several outdated npm packages. We identified the packages that we thought were safe to update to latest in `package.json`. After an npm install, tests and build both still pass.

- [x] Update `package.json` (see commit for details)
- [x] All tests pass
- [x] Build ... builds. ;)

A few notes:
- The update includes 1 top level package (promise) and the others are all dev dependencies.
- The update to promise simple changes it from `7.0.0.` to `^7.0.4`
- As we are not intimately familiar with the project, more eyes are needed on this to ensure everything works as expected.
- We did not update packages that were related to the test suite (Chai, Mocha, etc.) as doing so prevented tests from even running.

See justification for this update these changes here https://www.bithound.io/github/bithound/falcor/npm.updates

PR #536, opened 2015-09-09
Cleaned up package.json deps

sdesai:
* Updated dep versions (starting with https://github.com/Netflix/falcor/pull/468), and adding to it.
* Removed unused deps, and made wildcards consistent (^ for devDeps, fixed versions for product deps)
* Tested all npm run tasks, and main gulp tasks.

This PR replaces https://github.com/Netflix/falcor/pull/468 so will close that one out.


lib/get/walkPath.js
```diff
@@ -24,7 +24,7 @@ module.exports = function walkPath(model, root, curr, path,
     var keySet, i;
     keySet = path[depth];
 
-    var isKeySet = typeof keySet === 'object';
```
michaelbpaulson: My bad

michaelbpaulson:
+1

Issue #535, opened 2015-09-08
MaxRetryExceededError when a route returns a null value not wrapped in an atom


noonian:
I get this error if I request a route that is returning a null value (path exists and value for prop is null):

```js
Uncaught MaxRetryExceededError: The allowed number of retries have been exceeded.
```

The error is not thrown if the null is explicitly wrapped by the route in an atom. Errors are also not thrown for other falsey values (0, false, and ''). 

I believe this is a bug in the falcor client code, since there are [tests for the router that suggest it is legal it return null values](https://github.com/Netflix/falcor-router/blob/master/test/unit/core/get.spec.js#L29).

Here is a basic router and corresponding model invocation to reproduce the error:

```js
// index.js
var falcorExpress = require('falcor-express');
var Router = require('falcor-router');
var express = require('express');
var app = express();

app.use('/model.json', falcorExpress.dataSourceRoute(function (req, res) {
  return new Router(
    [
      {
        route: ['thing', 'prop'],
        get: function (path) {
          return {
            path: ['thing', 'prop'],
            value: null
          }
        }
      }
    ]);
}));

// statically host all files in current directory
app.use(express.static(__dirname + '/'));

var server = app.listen(3000);
```

client code:
```js
var model = new falcor.Model({source: new falcor.HttpDataSource('/model.json')});
model.get(['thing', 'prop']).then(function (res) { console.log(res); })
```

michaelbpaulson:
I can take a look at this, as it would make an excellent integration test. :)

michaelbpaulson:
hello @noonian I think I have found the problem.  What version of Falcor Router are you using?  That issue is fixed in [0.2.9](https://github.com/Netflix/falcor-router/tree/v0.2.9).

noonian:
Good morning,

I still see the issue with 0.2.9 with fresh downloads from npm. I don't think the issue is with the router though, since the server seems to be sending down the correct payload to the browser.

```js
{
  jsonGraph: {
    thing: {
      prop: null
    }
  }
}
```

So I think the problem is in the client side code that takes the router response and merges it into the cache (and presumably checks to see if it needs to do more requests).

Here is the package.json for the test project I'm using. I'm using falcor.browser.js hosted from netflix.github.io (pretty much the readme example from the router).

```json
{
  "name": "falcor-fun",
  "version": "0.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
[...]

PR #537, opened 2015-09-09
merge jsonGraph with a null non sentinel value.

michaelbpaulson:
Made a small change to merge jsonGraph node so that if the message and the cache are equivalent and nullsy then there are two outcomes, not one.  Now the merge will consider a `null` value differently than an `undefined` value.


michaelbpaulson:
#535

PR #538, opened 2015-09-09
Set pathMap local should perform follow-up get with requestedPaths to build proper JSON result.

trxcllnt:
@jhusain @michaelbpaulson Before, setting a PathMap on a local model where the PathMap spans existing references would build output JSON for the `optimized` paths, not the PathMap's `requested` paths. This PR fixes that.

### DIFF
--- a/test/set/jsonGraphs/expired.spec.js
+++ b/test/set/jsonGraphs/expired.spec.js
@@ -130,7 +130,7 @@ describe("an expired value", function() {
             }]
         );
 
-        expect(optimizedPaths.length).to.equal(0);
+        expect(optimizedPaths[1].length).to.equal(0);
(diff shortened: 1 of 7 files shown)


test/set/jsonGraphs/expired.spec.js
```diff
@@ -130,7 +130,7 @@ describe("an expired value", function() {
             }]
         );
 
-        expect(optimizedPaths.length).to.equal(0);
```
michaelbpaulson: Could we rename this from `optimizedPaths` to `successfulPaths`?  It can be a bit confusing.

trxcllnt: @michaelbpaulson I like it. will do.

PR #544, opened 2015-09-11
Swiftness

michaelbpaulson:


lib/get/getCachePosition.js
```diff
@@ -0,0 +1,31 @@
...
+
+    // The loop is simple now, we follow the current cache position until
+    //
+    while (++depth < maxDepth &&
+           currentCachePosition && !currentCachePosition.$type) {
+
+        currentCachePosition = currentCachePosition[path[depth]];
+    }
+
+    return [
```
sdesai: Do we need to return an array here? Expensive plus not very ergonomic. Can't we return `(null || currentCachePosition)` - or some other constant to reflect SHORTED. As far as I can tell, the consuming code doesn't need the currentCachePosition for the SHORTED case, and the CACHE_POSITION case doesn't need to know we're shorter.

sdesai:
One comment on the implementation (inline). Feel free to merge once addressed, since I don't have any more context on the flow than you do.

PR #546, opened 2015-09-12
Found a bug with how requested and optimized paths are reported.

michaelbpaulson:


michaelbpaulson:
@sdesai Here is the optimized and requested path fix for JSONGraph.

PR #547, opened 2015-09-12
Better core get

michaelbpaulson:
@sdesai This should be obvious that this is just a simple way to be more explicit with get tests.  I thought the other way may be easier to debug, but it was not.  This should be the right balance of debugability of get core and explicitness.


michaelbpaulson:
@sdesai this should be merged in after the optimized path fix.

PR #548, opened 2015-09-13
Error null bug

michaelbpaulson:
Found a small bug in how we are reporting errors when found at the end of a reference.

### DIFF
--- a/lib/get/onValueType.js
+++ b/lib/get/onValueType.js
@@ -52,7 +52,7 @@ module.exports = function onValueType(
     // If there is an error, then report it as a value if
     else if (currType === $error) {
         if (fromReference) {
-            requestedPath.push(null);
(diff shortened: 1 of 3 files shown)


michaelbpaulson:
@sdesai this uses the #547 as a base so I don't have to redo the test once merged in.  So merge #547 first before this one.

lib/get/followReference.js
```diff
@@ -55,7 +55,7 @@ function followReference(model, root, nodeArg, referenceContainerArg,
                 if (type === $ref) {
                     if (isJSONG) {
                         onValue(model, next, seed, null, null, null,
-                                reference, isJSONG);
```
sdesai: I assume this set of changes goes away, once merged with latest master

PR #551, opened 2015-09-18
New deref

michaelbpaulson:
The new version of deref.  @sdesai @jhusain


michaelbpaulson:
@sdesai we could gain some serious coverage if we deleted unused code, but I have avoided doing that for now.

trxcllnt:
Oh man, I need this asap. @michaelbpaulson I see `derefSync` still being used in `CallResponse`, are you saying we don't need to do that anymore?

michaelbpaulson:
@trxcllnt yes.  But We are keeping call as is until we remove local function calls (I know right...)

trxcllnt:
@michaelbpaulson ... sure.

trxcllnt:
@michaelbpaulson anything I can do to help move this PR along?

jhusain:
@michaelbpaulson

Should we be concerned that some checks have failed on Travis?

michaelbpaulson:
@jhusain its not travis, its coverage.  The coverage has dropped in `derefSync` and `getValueSync`.  But that is expected since those are going away here shortly.

sdesai:
Looks good to me. I had a couple of minor performance tweaks, which I'll mention inline, but nothing critical.

I did have a couple of more overarching questions:

a. Can you comment on the theoretical performance assumption.

The way I'm reading it, and based on what we discussed, there's some impact on get, but nothing substantial, since the main cost of maintaining the branchInfo per depth, is the object-per-depth creation cost, and we're re-using that object to replace the branch object we'd create onValue anyway, so should just be a couple of extra property sets.

I can run it through the performance tests also next week.

b. Did we decide on whether or not we want to bump up the major semver for backwards compat breaking changes like this?

lib/deref/index.js
```diff
@@ -1,61 +1,57 @@
...
-var pathSyntax = require("falcor-path-syntax");
+var __key = require("./../internal/key");
+var __path = require("./../internal/path");
+var __parent = require("./../internal/parent");
+var InvalidDerefInputError = require("./../errors/InvalidDerefInputError");
 
-module.exports = function deref(boundPathArg) {
+module.exports = function deref(boundJSONArg) {
+    var path = boundJSONArg && boundJSONArg[__path];
+    var key = boundJSONArg && boundJSONArg[__key];
```
sdesai: We could move the cost of this lookup to after the early path return (the invalid input error would need to be moved down also), but gives the fastest path out of deref.

lib/get/onValue.js
```diff
@@ -100,10 +102,15 @@ module.exports = function onValue(model, node, seed, depth, outerResults,
         if (!curr) {
             curr = seed.json = {};
         }
+        var branchValue;
         for (i = 0; i < depth - 1; i++) {
             k = requestedPath[i];
+            branchValue = branchInfo[i];
```
sdesai: Could move this inside the if(!curr[key]) check.

lib/deref/index.js
```diff
@@ -1,61 +1,57 @@
...
+    // get.  * - Technically this can just be passed in but don't do that.
+    if (path === undefined && key === undefined) {
+        throw new InvalidDerefInputError();
     }
 
-    if (modelRoot.syncRefCount <= 0 && pathsCount === 0) {
-        throw new Error("Model#deref requires at least one value path.");
+    // The simple deref when there is a path.
+    if (path) {
+        return this._clone({
```
sdesai: Optimizing clone was still pending right?

trxcllnt: @sdesai I think we already optimized clone? though there may be things we can do further, I'm not sure.

michaelbpaulson: There may be some things we can do to make clone faster but that is orthogonal to this PR.  I would like to measure that first, to confirm suspicions, before proceeding with optimizations.

michaelbpaulson:
All issues were addressed, going to merge.

jhusain:
@sdesai i see no reason not to rev semver for this change.

jhusain:
@sdesai Don't mean now, but whenever we decide it's appropriate to release.

michaelbpaulson:
Ideally @jhusain & @sdesai it would be great to rev semver when we remove Rx and path syntax from the core.  The path syntax should be a quick removal and so should rx at this point (since local function calls and deref have been dealt with).  The only small amount of work is around `SetResponse`.

But we don't have to couple those together either.  We could rev to 1.0.0 shortly then go to 2.0.0 soon after.

PR #552, opened 2015-09-18
No more local function calls and rewrote CallResponse to be more simple.

michaelbpaulson:
Now call responses are simple since they only need to forward the request to the source.  If the source does not exist then we will observer.onError function does not exist.

### DIFF
--- a/lib/response/CallResponse.js
+++ b/lib/response/CallResponse.js
@@ -41,9 +41,16 @@ function subscribeToResponse(observer) {
     }
 
 
-    var response;
-    return model._source.
-        call(boundCallPath, callArgs, suffixes, extraPaths).
+    var response, obs;
+    try {
+        obs = model._source.
+            call(boundCallPath, callArgs, suffixes, extraPaths);
+    } catch (e) {
+        observer.onError(new InvalidSourceError(e));
+        return;
+    }
+
+    return obs.
         subscribe(function(res) {
             response = res;
         }, function(err) {
@@ -60,12 +67,10 @@ function subscribeToResponse(observer) {
             rootModel.
                 withoutDataSource().
                 set(response).subscribe(function(x) {
-                    debugger
                     observer.onNext(x);
                 }, function(err) {
                     observer.onError(err);
                 }, function() {
-                    debugger
                     observer.onCompleted();
                 });
         });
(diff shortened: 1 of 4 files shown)


michaelbpaulson:
@sdesai the failures are due to sync methods getting lower coverage.  This is because they will be removed shortly all together, but I didn't want to pollute this PR with those removals.

michaelbpaulson:
@sdesai This is the merge that is the precursor to Rx becoming a devDependency.  If there is anything else that I need to do, let me know.  Remember, if you look at the coveralls drop its around some sync functions which still need to be removed.

sdesai:
Sorry about the delay. I've set aside some time this week to catch up on PRs, open issues. Look for some feedback, and/or this being merged tonight/tomorrow latest.

lib/response/CallResponse.js
```diff
@@ -9,241 +9,71 @@ var pathSyntax = require("falcor-path-syntax");
...
-    var args = this.args;
+    var callPath = pathSyntax.fromPath(this.callPath);
+    var callArgs = this.args;
+    var suffixes = this.suffix;
+    var extraPaths = this.paths;
     var model = this.model;
-
-    var callPath = pathSyntax.fromPath(args[0]);
-    var callArgs = args[1] || [];
-    var suffixes = (args[2] || []).map(pathSyntax.fromPath);
```
sdesai: Do we no longer expect path syntax for suffixes, extraPaths, or are we normalizing somewhere else?

sdesai:
Looks good to me. Just the one comment inline. Also is this considered back-compat breakage? Did we ever support local call functions formally (docs etc.)?

Feel free to merge, once that's addressed/resolved.

lib/response/CallResponse.js
trxcllnt: @sdesai it's not there, but it'd be nice if it were.

sdesai: I can re-add normalization, with tests, if that's the expectation.

trxcllnt: @sdesai since it was there before, removing it is a non-backwards compatible change. I think we should have it.

falcor-build: It's required. I'll implement it and add a unit test for it.

Issue #560, opened 2015-09-28
New deref from an already deref'd model doesn't include the parent model's path.


trxcllnt:
@michaelbpaulson @sdesai @jhusain I don't have time to submit a PR right this second, but [this line](https://github.com/Netflix/falcor/blob/master/lib/deref/index.js#L56) needs to be:
```es6
    _path: this._path.concat(nextPath)
```

This comes up if you do:
```es6
root.get('key1.key2.leafValue')
  .flatMap(({json}) => {
    var model1 = root.deref(json.key1);
    return model1.get('key2.leafValue').map(({json} => model1.deref(json.key2))
  })
  .subscribe((model2) => {
    console.log(model2._path); // will be ['key2'] instead of ['key1', 'key2']
  });
```

Obviously contrived example, but you get the point. Since the second get's JSON is relative to "key1", calling `deref` with the result of the second get doesn't include `key1` like it should.

PR #561, opened 2015-09-29
fixed the sequential deref bug.

michaelbpaulson:


michaelbpaulson:
#560

lib/deref/index.js
```diff
@@ -51,6 +51,13 @@ module.exports = function deref(boundJSONArg) {
         nextPath[++nextPathLength] = reversedKeys[i];
     }
 
+    // This is the key for creating a path from key/parent combo's only
+    // and there was already a bound path.  If we follow a reference (path
+    // variable is defined) then we do not need to concat any paths.
+    if (!path && this._path.length) {
```
sdesai: We should add a test case for the reference case also [ the `if (path && this._path.length)` case ]. I can add that.

PR #562, opened 2015-09-30
Removed inadvertent it.only

sdesai:
Was the cause of the reduced test coverage from https://github.com/Netflix/falcor/pull/552#event-422622779


michaelbpaulson:
Oops! Programming and merging with kids hours well :)

PR #566, opened 2015-10-07
CI will build now build *all* dists, and commits (to master)

sdesai:
So devs don't have to remember to build and commit dists with pull requests.


michaelbpaulson:
I love this.

sdesai:
It's my fabulous command of the English language in the title isn't it?

sdesai:
Merged. But it seems to have some env/authentication issue in the Travis CI env so it's not committing, plus may need to pick up the branch from CI env, since the build is off a detached HEAD. I'll look at that tomorrow.

sdesai:
Seems like it's fixed. Let me know if you see anything out of line

PR #567, opened 2015-10-07
Updating package.json to reflect latest published version

sdesai:
0.1.13 was published off a branch (since master already had back-compat breaking changes).

The package.json was updated for the publish, on the branch, but needs to be reflected in master, until we publish the next bump.


jameswomack:
:clap:

PR #569, opened 2015-10-09
Try committing the dist, one more time

sdesai:
1. Get off detached head, and onto the branch we're building (master in this case, but driven by $TRAVIS_BRANCH)
2. Add a new remote, same repo we're in, but using a url format Travis can push to.

### DIFF
--- a/build/deploy-ghpages.sh
+++ b/build/deploy-ghpages.sh
@@ -7,6 +7,7 @@ if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
   FALCOR_DOCS_DIR=$TEMP_DIR/falcordocs
   FALCOR_BUILD_DIR=$TEMP_DIR/falcorbuild
   GH_PAGES_DIR=$TEMP_DIR/gh-pages
+  DEPLOYABLE_REPO=https://${GH_TOKEN}@github.com/Netflix/falcor.git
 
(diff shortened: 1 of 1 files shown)


sdesai:
Does someone with CI/Bash skills want to lend a second set of eyes? Otherwise I'll merge it in tomorrow morning.

It's trying to fix this:
```
[detached HEAD 59e9e1d] Travis build 169 committed dist/
 4 files changed, 3175 insertions(+), 2929 deletions(-)
 rewrite dist/falcor.all.min.js (95%)
 rewrite dist/falcor.browser.min.js (64%)
```

And this, while attempting a `git push`

```
warning: push.default is unset; its implicit value is changing in
Git 2.0 from 'matching' to 'simple'. To squelch this message
and maintain the current behavior after the default changes, use:
  git config --global push.default matching
To squelch this message and adopt the new behavior now, use:
  git config --global push.default simple
See 'git help config' and search for 'push.default' for further information.
(the 'simple' mode was introduced in Git 1.7.11. Use the similar mode
'current' instead of 'simple' if you sometimes use older versions of Git)
remote: Invalid username or password.
fatal: Authentication failed for 'https://github.com/Netflix/falcor.git/'
```

sdesai:
Seems to work fine. Let me know if anyone sees anything out of line.

Issue #573, opened 2015-10-14
Observable and CompositeDisposable  declared but not used


HenriqueLimas:
In the [CallResponse.js](https://github.com/Netflix/falcor/blob/master/lib/response/CallResponse.js#L2) the ```Observable``` and the ```CompositeDisposable``` are declared but not used. It seems strange, is that correct?

falcor-build:
Wii Wii, I think our eslint isn't catching such simple errors!

HenriqueLimas:
The ```gulp lint``` is working but ```gulp all```, used in the ```dist``` script in the ```package.json```, is not calling the ```lint``` task.

PR #574, opened 2015-10-14
fix lint errors and add the lint task into build process.

HenriqueLimas:
Related to #573

### DIFF
--- a/lib/response/CallResponse.js
+++ b/lib/response/CallResponse.js
@@ -1,13 +1,7 @@
-var Rx = require("rx/dist/rx") && require("rx/dist/rx.aggregates");
-var Observable = Rx.Observable;
-var CompositeDisposable = Rx.CompositeDisposable;
-
 var ModelResponse = require("./../response/ModelResponse");
 var InvalidSourceError = require("./../errors/InvalidSourceError");
 
 var pathSyntax = require("falcor-path-syntax");
-var derefSync = require("./../deref/sync");
-var $ref = require("./../types/ref");
 
 function CallResponse(model, callPath, args, suffix, paths) {
     this.callPath = pathSyntax.fromPath(callPath);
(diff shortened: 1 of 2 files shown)


lib/response/CallResponse.js
```diff
@@ -41,8 +35,7 @@ function subscribeToResponse(observer) {
     // Precisely the same error as the router when a call function does not
     // exist.
     if (!model._source) {
-        observer.onError(new Error('function does not exist'));
-        return;
+        return observer.onError(new Error("function does not exist"));
```
sdesai: I think we should leave this as is. It's just more explicit that it returns "undefined" if it's on a separate line.

I was thinking about just adding an eslint comment override, to consciously acknowledge that we're returning 2 different values, and we're ok with it.

lib/response/CallResponse.js
```diff
@@ -1,13 +1,7 @@
-var Rx = require("rx/dist/rx") && require("rx/dist/rx.aggregates");
```
sdesai: This was the other area where I wanted to do some due diligence before pulling this out. We may need to go back and look at the history. I think this may have been added fairly recently to work around a build issue with WebPack/rx.aggregates. If not, this is fine to remove.

sdesai:
Thanks, I was about to add this, when I changed the build to 'all'. I realized we weren't linting, and hit some of the issues in CallResponse.js. A couple of comments inline, but otherwise seems fine to merge.

lib/response/CallResponse.js
HenriqueLimas: Yes, before I remove that I thought that maybe this is required for webpack load the rx files. But after I saw that the Rx is only used to load ```Observable``` and ```CompositeDisposable``` and CallResponse is not a Subclass of Observable anymore due to #552.

sdesai: Cool, that makes sense, thanks.

lib/response/CallResponse.js
sdesai: `/*eslint consistent-return: 0*/`

PR #577, opened 2015-10-14
Actually GC cache items when we collect the LRU.

### LINKED ISSUE #507: Collect does not adjust cache size.
The two places in which we call collect (in the upcoming getRequestCycle.js and IdempotentResponse.js) we do not pass in a version flag.  That version flag is required to update the graph which would update the size.  This is extremely problematic because every request to a dataSource will cause the LRU to be further cleaned up.

A good unit test would be to test lru/collect to ensure size changes.  Test a get that will exceed the cache and after the observable completes (some sort of setTimeout) the cache size must be reduced.

trxcllnt:
Pass the cache root's version to LRU collect after the get and set operations have finished mutating the cache. Fixes #507. cc: @michaelbpaulson @sdesai @jhusain

### DIFF
--- a/lib/response/IdempotentResponse.js
+++ b/lib/response/IdempotentResponse.js
@@ -9,6 +9,7 @@ var getSize = require("./../support/getSize");
 var collectLru = require("./../lru/collect");
 
 var arrayClone = require("./../support/array-clone");
+var __version = require("./../internal/version");
 
 var isArray = Array.isArray;
 var isPathValue = require("./../support/isPathValue");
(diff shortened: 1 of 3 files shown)


sdesai:
Sorry about the delay.

This looks fine to me, although I did want to verify I'm understanding `updateNodeAncestors` correctly:

It seems like we're relying on it to `removeNode`, with this change, and it will, starting at the sentinel/leaf (the first iteration of the do/while loop, when child.$size === offset) and then we'd continue up, either removing branches as their sizes are zero'd out, or updating version references (and size) if they're not.

Also, what's the use case for calling collectLRU without a version?

@michaelbpaulson - If you have a chance, can you look it over, since you filed the original issue. 

I'm thinking of cherry picking it into the 0.x branch, to pull it into the next patch release I'm pushing to npm.

michaelbpaulson:
+1

sdesai:
Just a heads up: Had to tweak the test a little when cherry-picking to 0.x, since 0.x didn't need to strip the internal deref keys before asserting [ 0.x has the 0.x deref ]

Issue #586, opened 2015-10-24
Webpack


IwanKaramazow:
What's the current status of the webpack issue?

I installed rx trough npm and added the following code to my webpack config
```
alias: {
      // Workaround https://github.com/Reactive-Extensions/RxJS/issues/832, until it's fixed
      'rx$': require.resolve('rx/dist/rx') 
    }
```
I get a:
```
ERROR in ./~/falcor/~/rx/dist/rx.aggregates.js
Module not found: Error: Cannot resolve module '12' in /pathtoproject/node_modules/falcor/node_modules/rx/dist
 @ ./~/falcor/~/rx/dist/rx.aggregates.js
```

Pointing the rx$ alias to 'falcor/node_modules/rx/dist/rx' doesn't solve the problem either:
=> webpack builds, but I get:
```
TypeError: this is undefined //firefox-console.
```

Is this an issue with Falcor, or did I mess something up?

IwanKaramazow:
Update:

Tried the https://github.com/Netflix/falcor/issues/428 approach, didn't work with latest version of Rx
```
=> Clone Falcor
webpack --output-library falcor browser.js falcor.webpack.js

//use falcor.webpack.js in my app gives me
TypeError: "name" is read-only //firefox-console
```

In source this refers to:
```
	InvalidSourceError.prototype = new Error();InvalidSourceError.prototype.name = NAME;InvalidSourceError.name = NAME;InvalidSourceError.is = function(e){return e && e.name === NAME;};module.exports = InvalidSourceError; /***/}, /* 26 */function(module,exports,__webpack_require__){module.exports = {iterateKeySet:__webpack_require__(27),toTree:__webpack_require__(28),toTreeWithUnion:__webpack_require__(29),pathsComplementFromTree:__webpack_require__(30),pathsComplementFromLengthTree:__webpack_require__(32),hasIntersection:__webpack_require__(31),toPaths:__webpack_require__(33),collapse:__webpack_require__(34)}; /***/}, /* 27 */function(module,exports){var isArray=Array.isArray; /**
```

sdesai:
Hey @IwanKaramazow 

We're in the process of removing the internal falcor dependency on Rx so there isn't really any value in upgrading to the latest version of Rx, which has the fix for this webpack issue (there's a migration cost, since it's a major bump in the rx version and signature changes), so you'd need to stick with the workaround for Rx.

* Are you just building falcor with webpack? It already has a webpack.config 
[...]

PR #591, opened 2015-10-25
Removed [fn].name override. Causes TypeError in strict mode

sdesai:
See: https://github.com/Netflix/falcor/issues/586#issuecomment-150840423

function.name is spec'd as read-only, and setting it, with "use strict" throws a TypeError in browser in which it's supported. In the browsers which support it, it's already "set", so we don't need to reset it.

IE seems to be the only outlier (haven't tested Edge): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#Browser_compatibility

If we do need to set it for some reason, we probably need to try/catch or feature test before setting it, but not sure if we really need it set on the constructor function.

### DIFF
--- a/lib/errors/InvalidModelError.js
+++ b/lib/errors/InvalidModelError.js
@@ -17,7 +17,6 @@ function InvalidModelError(boundPath, shortedPath) {
 // instanceof will be an error, but stack will be correct because its defined in the constructor.
 InvalidModelError.prototype = new Error();
 InvalidModelError.prototype.name = NAME;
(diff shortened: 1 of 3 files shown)


sdesai:
Closing this, and opening a PR against 0.x which is a branch I just pushed, off of the last 0.x publish from master. We can use the 0.x branch for minor/patch fixes since master is effectively 1.x now (it has the back-incompat deref change).

PR #598, opened 2015-10-28
Falcor npm version bump to pick up "use strict" fixes

sdesai:
There have been issues around leaked globals from falcor-path-syntax and function.name in falcor which prevents people from using falcor in a "use strict" environment.

I'm publishing a new version of falcor@0.1.14 which picks up these changes, so that falcor@0.1.14 should be usable with "Babel" and other environments which enforce "use strict".

Just issuing the PR for visibility, before I publish to npm, since there were a couple of modules which needed bumping, due to the falcor-paht-syntax bump.

  * falcor to 0.1.14
  * falcor-router to 0.2.11
  * falcor-json-graph to 01.1.7

Related threads:
https://github.com/Netflix/falcor/issues/586
https://github.com/Netflix/falcor-path-syntax/pull/25
https://github.com/Netflix/falcor-path-syntax/pull/21
https://github.com/Netflix/falcor/pull/597


sdesai:
Published. Let me know if you see anything out of whack.

Issue #589, opened 2015-10-24
Path returned from falcor-router is undefined after call


ekosz:
I'm using falcor-router 0.2.9 and falcor 0.1.13

Given I have a router that looks like this:

```js
{
  route: 'my',
  get: function my() {
    const userId = this.userId;
    return Promise.resolve().then(() => {
      if (!userId) {
        return { path: ['my'], value: $atom(userId) };
      }

      return {
        path: ['my'],
        value: $ref(['usersById', userId]),
      };
    });
  },
}, {
  route: 'auth.logout',
  call: function authLogout() {
    if (!this.userId) return [ { path: ['my'], invalidate: true } ];

    return oauthService.expireOldTokens(this.userId).then(() => {       
      this.res.clearCookie('authToken');
      return [
        { path: ['my'], invalidate: true },
      ];
    });
  },
}
```

When I run

```js
model.call('auth.logout')
```

I get the error

```
TypeError: Cannot read property 'length' of undefined
    at walkPath (node_modules/falcor/lib/get/walkPath.js:17:56)
    at innerGet (node_modules/falcor/lib/get/get.js:55:13)
    at checkCacheAndReport (node_modules/falcor/lib/response/get/checkCacheAndReport.js:28:19)
    at Observable._subscribe (node_modules/falcor/lib/response/get/GetResponse.js:60:19)
    at Observable.subscribe (node_modules/falcor/lib/response/ModelResponse.js:144:29)
```

I'm still trying to get my head around `call` statements and I'm struggling a bit. I have another route `auth.login` that is working just fine, I don't know why this one is failing the way it does.

jhusain:
Like it may be expecting an arguments array as a second parameter. We should change things to make this optional. For now try passing an empty array as the next parameter.

ekosz:
@jhusain I tried that, no luck. Looking at where the code is failing it looks like:

```js
// If there is not a value in the current cache position or its a
// value type, then we are at the end of the getWalk.
if ((!curr || curr && curr.$type) || depth === path.length) {
```

Where `path` is the 4th argument to the `walkPath` method.

sdesai:
I believe this is the same issue mentioned here:
https://github.com/Netflix/falcor/issues/477#issuecomment-152336762

Your envelope coming back (if you look at the network tab) probably looks something like:

```
{
   jsonGraph: {}
   paths: [],
   invalidated: [ ['my'] ]
}
```

Which falcor is not currently handling properly. Could you verify that?

[...]

Issue #477, opened 2015-08-24
"Cannot read property '$size' of undefined" after call()


dallonf:
In my app, I have a `orgs.list.push` `call` route. It returns the following:

```
return [
      {path: ['orgs', 'list']},
      {path: ['orgs', 'list', 'push'], value: $atom({ id: id })}
    ];
```

Which I assume (docs on both `call` and the Router are very sparse at the moment) invalidates the entire `orgs.list` path and returns a value `{id: id}`. (I'm returning a value at the path of the method itself in order to return arbitrary data related to the operation. Is this the way you're supposed to that?)

This works pretty well, as long as there's _already_ some data under `orgs.list` in the cache. If not, it throws this error (I had to include line numbers manually due to some odd sourcemap configuration in my project):

```
TypeError: Cannot read property '$size' of undefined
  at onNode (falcor/lib/invalidate/invalidate-path-sets-as-json-dense.js:102)
  at walkPathSet (falcor/lib/walk/walk-path-set.js:109)
  at walkPathSet (falcor/lib/walk/walk-path-set.js:112)
  at invalidatePathSetsAsJsonDense (falcor/lib/invalidate/invalidate-path-sets-as-json-dense.js:45)
  at InvalidateResponse.subscribeToInvalidateResponse [as _subscribe] (falcor/lib/response/InvalidateResponse.js:36)
  at InvalidateResponse.Rx.Observable.observableProto.subscribe.observableProto.forEach (rx/dist/rx.js:1976)
  at observableProto.finally.observableProto.ensure (rx/dist/rx.js:4242)
  at tryCatcher (rx/dist/rx.js:567)
  at setDisposable [as action] (rx/dist/rx.js:5233)
  at ScheduledItem.invokeCore (rx/dist/rx.js:929)
```

(Update: I tried commenting out the `{path: ['orgs', 'list', 'push'], value: $atom({ id: id })}` line, since that's a somewhat suspect thing that I'm trying - but it didn't seem to help, so it's probably unrelated)

quazzie:
I have the same problem...

jhusain:
To invalidate the entire list you return a path-bound invalidation:
[
{path: ["orgs", "list"], invalidated: true }

]

quazzie:
@jhusain But if you invalidate a path that is not cached you get the $size error.
On line 2357 in falcor.browser.js it does not check if the node exists in cache.

trxcllnt:
[...]

PR #600, opened 2015-10-30
Fixes exceptions on call responses with only invalidations, and no returned paths

sdesai:
We ended up trying to make a get with a requestedPath = [undefined], due to a bug in arrayFlatMap which wasn't accounting for the selector filtering out values, by initializing the flatMapped array to a fixed size.

See #589
See #477#issuecomment-152336762

@michaelbpaulson @jhusain


michaelbpaulson:
+1

PR #608, opened 2015-11-06
Migrations

michaelbpaulson:


MIGRATIONS.md
```diff
@@ -0,0 +1,186 @@
...
+
+        // Other rendering stuff / application logic
+        ...
+    });
+```
+
+##### Pros
+* Simpler to grok/use
+* Promotes better application architecture.
+* Always syncronous.
```
chrisvariety: `synchronous` :wink:

MIGRATIONS.md
```diff
@@ -0,0 +1,186 @@
...
+            1337: {
+                title: Model.atom('Total Recall (June 1st, 1990)')
+            }
+        }
+    }
+});
+```
+
+If we were to use `deref` in the **old** way we would have to perform the
+following to derefernce to [genreLists, 0, 0].
```
chrisvariety: `dereference` :smile:

PR #611, opened 2015-11-09
Fixes errorSelector bugs.

sdesai:
a. errorSelector was not being invoked for branches/replaced nodes
   during merges. It was only invoked on new leaf insertion.

b. errorSelector was being called with truncated paths. This impacted
   both sets, and merges from the datasource response.

*NOTE*. This should not be merged yet. Unit tests coming. Wanted to get some eyes on the fix.

@michaelbpaulson


lib/support/reconstructPath
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @param  {Array} currentPath The current requestedPath state, during the walk
+ * @param  {String} key        The current key value, during the walk
+ * @return {Array} A new array, with the path which represents the node we're about
+ * to insert
+ */
+module.exports = function reconstructPath(currentPath, key) {
+
+    // The first time called during a walk, currentPath.index won't be set.
+    var currentPathIndex = (currentPath.index || 0) + 1;
```
michaelbpaulson: The only thing I don't like about this, and we will have to change it at some point, is that set relies an mutating the input with `index`.  We can rely on this for now.

sdesai: I agree. I'm happy to feed in index, but it seems like those are heavily coupled already (requestedPath and requestedPath.index always have to by in sync don't they?), so separating them here seems riskier, unless we separate them all the way up.

trxcllnt:
They can be separated, but yes they're always updated in lock-step. Index was added after the recursive function signatures were defined (I didn't realize we needed it till later in the game), so it was purely a cosmetic choice not to go back and update all the signatures (and call-sites) with index params.

sdesai:
Makes sense. Planning to leave them alone for the purposes of this fix/change to get errorSelector working properly. Can treat as a cleanup task.

sdesai:
Needed to take another stab at it, since due to the nature of the do/while loop, index wasn't enough info to reconstruct the errorPaths fed to the selector for PathSets with keySets of size > 1, e.g. `[foo, bar, [baz, fizz]]` or `[foo, [baz, fizz], bar]`. I needed to add a `depth` which would stay the same when processing both the `baz`, and `fizz` keys. Chose to add it to the requestedPath "object" (even though it was fairly trivial to add it as an extra arg, since that's effectively the datastructure we're using to carry around the requestedPath related state through the recursion and related method calls and needs to stay in sync.

Again, eyeballs appreciated.

sdesai:
@michaelbpaulson - Gonna merge this in, unless you you want to take a look at it?

michaelbpaulson:
Sounds great to me!

PR #612, opened 2015-11-10
0.x wip

michaelbpaulson:
@sdesai there is an interesting thing in here.  I noticed that `wrapNode` is called in different order so if the `errorSelector` returned anything but the value it would remove the whole value from the cache (size becomes 0), unless the `errorSelector` also provides the size.  By moving `wrapNode`, all is well in the universe.

### DIFF
--- a/lib/support/mergeJSONGraphNode.js
+++ b/lib/support/mergeJSONGraphNode.js
@@ -150,6 +150,7 @@ module.exports = function mergeJSONGraphNode(
 
         if (mType === $error && isFunction(errorSelector)) {
             message = errorSelector(reconstructPath(requestedPath, key), message);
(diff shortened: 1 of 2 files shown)


sdesai:
Makes sense (James was asking about $size yesterday in general for the transforms he wanted to play around with - would wrapNode fix size also, if the node is already wrapped, but the value changed?).

I think we should think through the implications (does errorSelector need to be renamed, should all values pass through it - not just errors, and it's just a general transformValueForCache, what's the contract for what should be returned - value, atom, anything? etc.) and file an issue for this, if we're thinking of leveraging errorSelector for more than just error => error transforms (which seems like the original intent), and we can go from there in terms of moving it into the code base formally.

PR #613, opened 2015-11-11
mType instead of cType.

michaelbpaulson:


michaelbpaulson:
@sdesai this is correcting the `cType` to `mType` issue.

trxcllnt:
If `node` and `message` are the same, wouldn't their types also be the same? Or am I missing something.

michaelbpaulson:
@trxcllnt its an experimental branch attempting to solve an internal problem in a very short amount of time.  It allows for error selectors to change the _type_ of values going in and out.  Essentially, having a secondary fallback model that is invoked upon error and potentially able to give a value that has some value.

Its currently done in this side branch as its not the solution we want, its just the easiest solution from point a to b.

trxcllnt:
@michaelbpaulson oh, excellent. that sounds like something the errorSelector should be able to do.

PR #618, opened 2015-11-13
added custom logic to the default comparator to consider $expires.

michaelbpaulson:


michaelbpaulson:
@sdesai made the default comparator consider some meta data.

PR #629, opened 2015-11-25
Set add abs paths

michaelbpaulson:
@sdesai an unusual amount of files changed for this commit due to the following.

* There was a bug in get promote so I merged get promote with set promote and then made get / set use the same promote (prevent bugs). 
  * After finding this bug I added 2 unit tests to ensure we get behavior we want.
  * Fixing the LRU was required to ensure working behavior of new deref.

* We were not collecting the LRU on a sync get.  This is 99% not an issue until a set operation is performed that exceeds the cache's maximum size.  I would not of caught this issue without this feature since the testing uses this case to ensure that a collected reference looks a certain way (were _not_ manually deleting properties, which is true).

* We have the flags to twiddle the output generation for steve.

* We had a bad lint rule that allowed me to treat variables not as block scoped (it just warned so I boosted its warning to error and fixed the one location we were doing this).


michaelbpaulson:
@sdesai I have updated this PR.  You should take a look at it asap.  Then we could merge in this PR and call deref "done."

lib/Model.js
```diff
@@ -106,6 +106,11 @@ function Model(o) {
         this._treatErrorsAsValues = options._treatErrorsAsValues;
     }
 
+    this._allowEnumerableMetadata = options.allowEnumerableMetadata ||
```
sdesai: * Is there still a need for allowEnumerableMetadata? If we're going with the obfuscated path?
* Also, if we're still experimenting with "allowFromWhenceYouCame", should we keep it off the public options, and keep it _ private only.

lib/Model.js
```diff
@@ -569,6 +587,17 @@ Model.prototype.getPath = function getPath() {
     return arrayClone(this._path);
 };
 
+/**
+ * This one is actually private.  I would not use this without talking to
+ * jhusain, sdesai, or michaelbpaulson (github).
```
sdesai: I would add a warning out it potentially changing in future, and it being experimental

lib/deref/hasValidParentReference.js
```diff
@@ -0,0 +1,33 @@
+var __parent = require("./../internal/parent");
+var __invalidated = require("./../internal/invalidated");
+module.exports = function fromWhenceYeCame() {
+    var reference = this._referenceContainer;
```
sdesai: Worth clarifying what "this" is assumed to be in this context, since it's just a standalone function.

On the other hand, if this is not on a critical path (that is, it's not called 1000s of times for a typical use case), could we just have this accept a model, and pass it through from the Model.prototype._hasValidParentReference() instead, like the walk implementations.

It just seemed a little odd to see a standalone function, reference "this", when "this" can only really ever be a Model instance, or something which shares the Model interface.

I don't think we need to change this for this PR (since it looks like we do it in quite a few places for stuff under lib), but worth considering for future development/code organization.

lib/deref/hasValidParentReference.js
```diff
@@ -0,0 +1,33 @@
...
+
+    // Always true when this mode is false.
+    if (!this._fromWhenceYouCame) {
```
sdesai: Wasn't this a function? In Model line 595? Am I missing something? Is it supposed to be _allowFromWhenceYouCame?

lib/deref/index.js
```diff
@@ -1,65 +1,62 @@
...
-    // We have to follow the path back up recursively until we hit a path
-    // or a null parent.
-    var reversedKeys = [key];
-    var reversedLength = 0;
-    var current = boundJSONArg[__parent];
-    path = null;
-    while (current !== null && !path) {
-        key = current[__key];
-        path = current[__path];
+    var absolutePath = boundJSONArg && boundJSONArg.$__path;
```
sdesai: Did we decide to stop using the internal/prefix? Just want to make sure it's conscious, since I don't remember it being part of the __path obfuscation discussion.

lib/deref/index.js
```diff
@@ -1,65 +1,62 @@
...
+    if (absolutePath) {
+        var validContainer = CONTAINER_DOES_NOT_EXIST;
+
+        if (toReference) {
+            validContainer = true;
+            referenceContainer = getCachePosition(this, toReference);
+            // If the reference container is still a sentinel value then compare
+            // the reference value with refPath.  If they are the same, then the
+            // model is still valid.
+            if (refPath && referenceContainer && referenceContainer.$type) {
```
sdesai: Discussed this in person, but just for visibility/anyone else following the PR: Need to make sure validContainer = false, unless "cache has a $type = ref at the same toReference path, and that reference array is the same as refPath".

lib/get/followReference.js
```diff
@@ -59,7 +59,7 @@ function followReference(model, root, nodeArg, referenceContainerArg,
                         onValue(model, next, seed, null, null, null, null,
                                 reference, reference.length, isJSONG);
                     } else {
-                        promote(model, next);
+                        promote(model._root, next);
```
sdesai: This was the lru/reference promotion fix? Just checking.

lib/get/util/clone.js
```diff
@@ -7,7 +7,7 @@ module.exports = function clone(node) {
     outValue = {};
     for (i = 0, len = keys.length; i < len; i++) {
         var k = keys[i];
-        if (k[0] === prefix) {
+        if (k[0] === prefix || k === "$absolutePath") {
```
sdesai: We need the prefix check here, because we're still using the prefix on cache meta-data currently? Just not on the JSON payload meta-data?

lib/lru/promote.js
```diff
@@ -1,44 +1,47 @@
...
-            tail = root[__tail],
-            next = node[__next],
-            prev = node[__prev];
-
-        if (node !== head) {
+// [H] -> Next -> ... -> [T]
+// [T] -> Prev -> ... -> [H]
+module.exports = function lruPromote(root, object) {
+    // Never promote node.$expires === 1.  They cannot expire.
+    if (object.$expires === 1) {
```
sdesai: Did we get rid of the constant (from the expires-now module) for a reason?

sdesai:
Went through most of this. Feel free to merge once cleaned up.

PR #637, opened 2015-11-30
found that references were not being promoted.

michaelbpaulson:
@sdesai


sdesai:
Can't this just be a merge from 0.x? Just so the commit SHAs stay the same/history is maintained. Does it need to be a manually replicated commit?

PR #639, opened 2015-12-01
The core was not considering already materialized atoms correctly

michaelbpaulson:
@sdesai There is a get rule that requires if a path is missing `get` will only materialize it if there is a `_source` and `_materialized` is true, but if there already is a path that has been materialized, to report it only requires `_materialized` to be true.  I was using the former logic to generate the output, thus incorrectly generating empty output for most cases.


michaelbpaulson:
@sdesai when you get the chance review and merge.

Issue #601, opened 2015-10-31
model.getCache throws 'Maximum call stack size exceeded' error 


threepointone:
This usually happens after making a request to an HttpDataSource. I have a standalone reproducible case (pardon the es6/7) -

```js
import request from 'superagent';
import Router from 'falcor-router';
import falcor from 'falcor';

// a 'service' that fetches user details from github
function user(id){
  return new Promise((resolve, reject) =>
    request.get(`https://api.github.com/users/${id}`).end((err, res) =>
      err ? reject(err) : resolve(res.body)));
}

var model = new falcor.Model({
  source: new Router([{
    // define a single route corresponding to the above service
    route: 'users[{keys:ids}]',
    get: async function (pathSet){
      let responses = await Promise.all(pathSet.ids.map(user));
      return pathSet.ids.map((id, i) =>
        ({path: ['users', id],  value: responses[i]}));
    }
  }])
});

console.log(model.getCache()); // empty, all good

model.get(`users.threepointone['id', 'url', 'name']`)
  .subscribe(function(json){
    // throws error
    // Maximum call stack size exceeded
    console.log(model.getCache());
  }, function(error){
    console.error(error);
  });
```

I see that the problem is happening in a _copyCache function, where it's recursively going through its internal cache, but following a `parent` key (and others) resulting in a circular reference?

michaelbpaulson:
Hmm.  This should be an easy fix.  I will have to look into this shortly, if I am the one to fix it :)

threepointone:
thanks, I'll try today!

threepointone:
It looks like every object is tagged with a number of keys, could you explain how you prevent clashes with pre-existing keys? 
<img width="232" alt="screen shot 2015-11-27 at 6 18 31 pm" src="https://cloud.githubusercontent.com/assets/18808/11441569/6700bfc4-9533-11e5-9925-cb3c3e7517e5.png">

The code in question is this bit [here](https://github.com/Netflix/falcor/blob/master/lib/get/getCache.js#L21) - I'm unsure whether I should remove any key named `parent`, since objects could come in with said key. also, how to deal with `version`?

michaelbpaulson:
@threepointone Yes, those keys are actually something we were going to change anyways.  Essentially what is going on here is that parent key is actually <30>parent where <30> is ascii code 30, record separating character.  We are going to go with a $ one here for performance reasons, but for now you can always (which I thought it did) for the first character being that prefix character remove it for the depth recursing.

[...]

PR #640, opened 2015-12-01
getCache was still getting serializable issues

michaelbpaulson:
So instead of doing 
```
x[0] !== prefix && x !== '$size';
```
we switched it up with a more verbose check.

```
module.exports = function isInternalKey(x) {
    return x === __parent ||
        x === __key ||
        x === __version ||
        x === __prev ||
        x === __next ||
        x === "$size";
};
```
#601

### DIFF
--- a/lib/get/getCache.js
+++ b/lib/get/getCache.js
@@ -1,6 +1,6 @@
 var $modelCreated = require("./../internal/model-created");
 var clone = require("./util/clone");
-var prefix = require("./../internal/prefix");
+var isInternalKey = require("./../support/isInternalKey");
 
 /**
  * decends and copies the cache.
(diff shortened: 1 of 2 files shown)


lib/get/getCache.js
```diff
@@ -18,7 +18,7 @@ function _copyCache(node, out, fromKey) {
     Object.
         keys(node).
         filter(function(k) {
-            return k[0] !== prefix && k !== "$size";
```
trxcllnt: @michaelbpaulson @jhusain @sdesai this check used to catch all our "internal" keys (the ones with the [prefix](https://github.com/Netflix/falcor/blob/master/lib/internal/prefix.js)), but now internal keys (like back-references) are leaking into the result of `getCache` and confusing people :(

sdesai: I'm re-establishing this (just with charAt(0) instead of the index lookup, just to be safe), in the 0.x branch as we speak (we just hit it internally too - the refN, ref-index, etc. keys ended up in the cache).

Was just trying to collect context from @jhusain @michaelbpaulson on why we ended up going with the new impl in this PR, in 0.x, and adding unit tests, and will be ready for new PR tomorrow.

PR #644, opened 2015-12-02
Comparator upgrades to have more fine grain comparison

michaelbpaulson:
@sdesai this is for internal need.  I'll cherry-pick this commit along with #639 shortly.

### DIFF
--- a/lib/ModelRoot.js
+++ b/lib/ModelRoot.js
@@ -28,11 +28,14 @@ function ModelRoot(o) {
 ModelRoot.prototype.errorSelector = function errorSelector(x, y) {
     return y;
 };
-ModelRoot.prototype.comparator = function comparator(a, b) {
-    if (hasOwn(a, "value") && hasOwn(b, "value")) {
-        return a.value === b.value;
+ModelRoot.prototype.comparator = function comparator(cacheNode, messageNode) {
+    if (hasOwn(cacheNode, "value") && hasOwn(messageNode, "value")) {
+        // They are the same only if the following fields are the same.
+        return cacheNode.value === messageNode.value &&
+            cacheNode.$type === messageNode.$type &&
+            cacheNode.$expires === messageNode.$expires;
     }
-    return a === b;
+    return cacheNode === messageNode;
 };
 
 module.exports = ModelRoot;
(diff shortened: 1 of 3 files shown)


lib/ModelRoot.js
```diff
@@ -28,11 +28,14 @@ function ModelRoot(o) {
 ModelRoot.prototype.errorSelector = function errorSelector(x, y) {
     return y;
 };
-ModelRoot.prototype.comparator = function comparator(a, b) {
-    if (hasOwn(a, "value") && hasOwn(b, "value")) {
-        return a.value === b.value;
+ModelRoot.prototype.comparator = function comparator(cacheNode, messageNode) {
+    if (hasOwn(cacheNode, "value") && hasOwn(messageNode, "value")) {
```
sdesai: This seems fine for now if we want to merge.

I'm wondering about the $expires check though (the $type check makes complete sense).

If $expires is used to define "equality", then how about other meta-data which may be added to the wrapper? Say, $absolutePath in the new deref world if the only thing which changed for this atom is the absolute path to it, or if the user added their own meta-data to the wrapper, would they need to update the method to account for equality wherever this is used.

Basically just wondering why "expires" is part of "value" equality.

michaelbpaulson: Well its not value equality.  Its _distinct_ checking.  Are these two items distinct.  $type-values, at least after thinking about it, are default distinct on 3 items, $type, $expires, and value.  Think of the following situation.

```javascript
model.setValue(['path'], atom(5, {$expires: Date.now() + 1000}).subscribe();
model.setValue(['path'], atom(5, {$expires: Date.now() + 2000}).subscribe();
```

Without $expires checking the second set would never happen.

sdesai: Makes sense, but in that case, isn't there the potential to have other meta-data which impacts the  'distinct' comparison.

I guess what I'm trying to get at is, should we compare all meta-data ($) props, generically, by just iterating them (performance would be a reason not to), instead of picking out expires specifically. I could pass different $foo values in the meta-data arguments in your example above, which would make the nodes distinct.

michaelbpaulson: I also was going to do the same thing.  

* There is a major performance concern.  I can quantify the numbers but it will no be cheap.  
* Metadata outside our domain knowledge should probably be up to the user to compare.  They could provide their own comparator if they need a new/additional set of rules.

Imagine this
```javascript
var a = {
    $type: 'atom',
    $meta: { lots: 'of', fields: 'here'},
    value: 6
}
var b = deepCopy(a);
comparator(a, b) // with all meta keys compared, this would be false, since object comparison is always false if they are not the same object.
```

PR #645, opened 2015-12-02
Comparator and materializing

michaelbpaulson:
@sdesai this is the cherry picking of #644 #639


sdesai:
It's breaking unit tests currently.

michaelbpaulson:
@sdesai yes.  I was going to address this once the other related PRs are merged so that way I can "maintain" somewhat similar git log, until I need to diverge.

PR #659, opened 2015-12-09
0.x -> wip

michaelbpaulson:


michaelbpaulson:
I think something is broken as my PR worked and now its spinning its gears on travis commits...

PR #667, opened 2015-12-14
Fixed deref completing, without onNext'ing a bound model, if the preload paths fail

sdesai:
We were swallowing errors on preload paths, and not onNexting a successfully bound model as a result.

This PR aims to fix that (For the current 0.x deref. It's N/A for the new 1.x/master deref).

@michaelbpaulson - Could you take a quick look at this, including the unit tests (to make sure I'm testing the correct 'as designed' expectation).


michaelbpaulson:
boom.  Take that 1.

PR #671, opened 2015-12-17
Fix getCache() internal key filtering

sdesai:
I didn't get a chance to get the context around this PR from Michael in person before he stepped out on vacation: https://github.com/Netflix/falcor/pull/640/ but it seems to have regressed `model.getCache()` behavior, by allowing some internal keys to remain in the results of `model.getCache()` (e.g. refN, ref-index, etc). It seems to be a pretty conscious change, so not sure if it was intentional.

However, in the absence of this context, it seems like the desired behavior is to strip *ALL* internal keys from the cache, when returning a copy through `getCache()`.

This PR aims to fix the above regression, and also fix what seems to be a gap in the original implementation, where `$size` would still be left in the results of `getCache()` at the leaf/sentinel nodes, when cloning boxed values. I don't think this was intentional.

Unit tests added, to make sure:

1. Stripping `refN`, `ref-index` doesn't result in references breaking when the copy of the cache is used to re-hydrate a new model
2. Make sure no internal keys are left in the cache-copy returned by `getCache()`

@jhusain @michaelbpaulson @trxcllnt - eyes appreciated, especially since I'm making assumptions about the original design intention.

### DIFF
--- a/lib/get/getCache.js
+++ b/lib/get/getCache.js
@@ -1,5 +1,4 @@
 var $modelCreated = require("./../internal/model-created");
-var clone = require("./util/clone");
 var isInternalKey = require("./../support/isInternalKey");
 
 /**
@@ -11,6 +10,25 @@ module.exports = function getCache(cache) {
 
     return out;
 };
+
+function cloneBoxedValue(boxedValue) {
+    var clonedValue = {};
+
+    var keys = Object.keys(boxedValue);
+    var key;
+    var i;
+    var l;
+
+    for (i = 0, l = keys.length; i < l; i++) {
+        key = keys[i];
+
+        if (!isInternalKey(key)) {
+            clonedValue[key] = boxedValue[key];
+        }
+    }
+
+    return clonedValue;
+}
 
 function _copyCache(node, out, fromKey) {
     // copy and return
@@ -40,7 +58,7 @@ function _copyCache(node, out, fromKey) {
                 var isUserCreatedcacheNext = !node[$modelCreated];
                 var value;
                 if (isObject || isUserCreatedcacheNext) {
-                    value = clone(cacheNext);
+                    value = cloneBoxedValue(cacheNext);
                 } else {
                     value = cacheNext.value;
                 }
--- a/lib/support/isInternalKey.js
+++ b/lib/support/isInternalKey.js
@@ -1,22 +1,12 @@
-var __parent = require("./../internal/parent");
-var __key = require("./../internal/key");
-var __version = require("./../internal/version");
-var __prev = require("./../internal/prev");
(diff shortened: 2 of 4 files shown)


trxcllnt:
@sdesai 😍 thanks! cc: @ekosz

lib/get/getCache.js
```diff
@@ -40,7 +58,7 @@ function _copyCache(node, out, fromKey) {
                 var isUserCreatedcacheNext = !node[$modelCreated];
                 var value;
                 if (isObject || isUserCreatedcacheNext) {
```
sdesai: This was another part of the original implementation I didn't understand the motivation for.

I'm not sure what the cacheNext.value === 'object' check was trying to do. We didn't end up cloning cacheNext.value anyway (that is, we didn't deep clone the wrapper in this case).

It seems to imply we only want to maintain the wrapper, for values which are objects (or for nodes which are pre-wrapped by the user), and I don't understand the motivation for the 'values which are objects' decision.

michaelbpaulson:
@sdesai All is correct.  The motivation for 'values which are objects' decision is a subtle one.

```javascript
var m = new Model();
var m2 = new Model();

// set a complex object.
m.setValue(['foo'], {bar: 'baz'}).subscribe();

// Set the cache from m to m2
m2.setCache(m.getCache());
```

Without the check for objectyness the following would happen.
```javascript
m2.get(['foo']).subscribe(log, log, log.bind(null, 'i completed!'));
```
This would simply console out `i completed!` because the underlying cache changed from m to m2.

m and m2's cache would be
```
// m
{
    foo: {
        $type: 'atom',
        value: {
            bar: 'baz'
        }
    }
}

//m2
{
    foo: {
            bar: {
                $type: 'atom',
                value: 'baz'
            }
    }
}

```

Hope that helps!

sdesai:
I see. Thanks. Makes sense. It's to distinguish between branches and atoms, and we don't need to take on the cost of the wrapper/clone for the non-object case, because there's nothing ambiguous to distinguish against.

Gonna merge this.

trxcllnt:
@sdesai is this going to be merged into master soon?

sdesai:
Hey Paul, I believe master has the original (prefix) based filter already. Otherwise can merge it in. This one does have the additional fix for sentinel wrappers

trxcllnt:
@sdesai from what I can tell, master's version of [`isInternalKey`](https://github.com/Netflix/falcor/blob/52f601756222e210cb231f1ab4f02491a642863e/lib/support/isInternalKey.js) is still checking for specific internal keys.

sdesai:
Ok. I'm OOO for Xmas, but will merge it in, in between things

PR #676, opened 2015-12-26
Fixes bug in walkPath that could corrupt cache references …

trxcllnt:
… because they aren't defensively copied. Also defensively copies new internal `deref` references in the output JSON so they can't be corrupted by user code (this was the source of a bug for me recently). Normalizes internal key definitions.


michaelbpaulson:
I do not believe we are going to defensively copy that reference, for get performance reasons.  We denoted with a `$` identifying it as a reserved member of falcor and then with 2x `_` to make it _extra private_.

If you would like to muck with the internals then let your program copy the array since it can take on the extra running time cost.

As far as the internal key condensing, I really like that as we were going to do that here shortly anyways.  So if you would like to make a second PR for that I would be glad to merge.

trxcllnt:
Well regardless, we do have to defensively copy the reference values at least, so they can't get corrupted by user code (otherwise you get really difficult to track down bugs when _.merge writes unexpectedly to your Falcor cache).

Nothing in Falcor internals uses "$" prefixes, everything that does is user-configurable metadata. My argument for moving away from it in the output is to avoid confusing people who know how to use metadata.

I need the other Arrays copied ($__path and $__toReference) for the same reasons, but visiting every node in the output is prohibitively expensive. Perhaps we can explore other avenues for customizing the output JSON if copying the lists is indeed a performance issue (of which I'm not yet convinced).

trxcllnt:
@michaelbpaulson I'll rename this branch and revert the Array cloning, but before I do, would you like to keep the internal keys prefixed with our internal `prefix` value, or reverted to the `$__` prefix?

Additionally, would you be amenable to a commit which includes the option to call out to a user-provided function that would allow someone to customize the output of the internal `get` and `set` implementations (similar to what we already do with the `errorSelector`)? This would allow me to customize the `get` output to my liking (for example, using `Object.defineProperties` to ensure the private keys are non-enumerable) without impacting core `get` and `set` performance.

@jhusain @sdesai

trxcllnt:
@michaelbpaulson @jhusain @sdesai after reviewing further, I'm even more convinced we should *always* copy *all* the metadata required for the new synchronous `deref` implementation.

The `$__path` and `$__toReference` values are assigned from each node's `$__absolutePath`, which is only calculated when a node is *set* into the graph, and the `$__refPath` is a reference to the actual `value` Array of a `{ $type: 'ref' }` in the cache. Falcor makes a guarantee that mutating any part of the JSON output is 100% safe, since all branches/values are copied (even in boxValues mode, the actual atom wrappers are copies of the cache's atoms). Since these values not copied, anything as trivial as a `lodash.merge` inside a `scan` can corrupt the `$__absolutePath` and reference values.

michaelbpaulson:
We are not going to do this afaik.  @jhusain feel free to re-open if I am incorrect.

PR #677, opened 2015-12-27
Add onChangesCompleted callback to ModelRoot

trxcllnt:
This handler is similar to the "onChange" callback, but is invoked after all updates to the cache have finished, all output has been delivered, and all outstanding Observables have terminated with an error or completion.

@jhusain @sdesai


michaelbpaulson:
We are not going to do this afaik.  If I am incorrect @jhusain please feel free to re-open pr.

PR #685, opened 2016-01-05
feat(performance): Removing of dynamic keys from get

michaelbpaulson:
There is still plenty of room in set to continue this work but this is
the first step to get rid of dynamic keys.  The performance win of
stripping from LRU, HardLinking, and following references roughly made
a 20% performance improvement _on device_.


michaelbpaulson:
@steveorsomethin please be happier with me :)

PR #686, opened 2016-01-06
Attempt to merge 0.x into master

sdesai:
@michaelbpaulson - Attempting to merge 0.x into master, to pick up my `getCache` internal key fixes.

It looks like I'm picking up a couple of other deltas (if you look at the Files Changed):

a. My error selector fixes (https://github.com/Netflix/falcor/pull/611)
b. Some other fixes around lib/get/getValueSync, lib/deref/sync

Could you look at b. - see if they make sense.

As discussed, the `isInternalKey` impl will change anyway, with the perf-optimization you're working on, but figured it would be worth seeing if we could merge 0.x back into master in general, so it's easier to pick up these changes in 0.x, until we release 1.x.

NOTE: The only files I had to do a `git checkout --ours` (as in master) on were:

1. The dist/. files, of course
2. The lib/deref/index.js file (to pick the master impl, in it's entirety).

Other than that, I fixed a few merge conflicts, and fixed some tests (to account for the Rx removal, and other smaller issues around test cache data setup)

If it looks too scary, we can just cherry-pick, and close out this PR.


michaelbpaulson:
@sdesai Its x'd!!!

lib/support/isInternalKey.js
```diff
@@ -1,24 +1,13 @@
...
  */
 module.exports = function isInternalKey(x) {
-    return x === __parent ||
-        x === __key ||
-        x === __version ||
-        x === __prev ||
-        x === __next ||
-        x === "$size" ||
-        x === $absolutePath;
+    return (x === "$size") || (x === "$absolutePath") || (x && (x.charAt(0) === prefix));
```
michaelbpaulson: @sdesai here is the point of contention.  I believe prefix is now `$$`

sdesai: Right - I was assuming that change would come with your set of performance changes (as discussed yesterday). I was just making it work with the 'current' state of master code.

If your changes are in, I can make that change along with this request.

sdesai: I see they're in, with the merge conflict I have to fix also. I'll make the change

sdesai:
@michaelbpaulson - I fixed the merge conflict + updated isInternalKey to match the current choice of prefix. The drop in coverage is due to the changes in getValueSync being merged in, which I needed you to take a look at - to see if we need them or not, from 0.x

PR #687, opened 2016-01-07
Fix a bug in updateBackReferenceVersions that could skip updating a back-reference branch.

trxcllnt:


michaelbpaulson:
@trxcllnt nice!

michaelbpaulson:
@trxcllnt Arrg!!! I hit the wrong button before being done here.  You forgot to add a unit test to confirm this behavior / shed light onto what you fixed.

PR #689, opened 2016-01-07
Performance issue.

michaelbpaulson:
Somewhere along the way in 2 performance commits that yielded great results could not be realized on master.  In fact, a performance degrade happened.  I am going to "retry" this change and measure more often and in between to find out what happened.


michaelbpaulson:
@sdesai I am unable to track down where performance went sideways with this change so I am going to revert it and try again.  This will be a great time for me to go over the available performance tests and help create that full library of tests.

PR #690, opened 2016-01-07
feat(performance): The removal of extra LRU promotes.

michaelbpaulson:
There is definitely a performance win here.  When the extra promotes
are removed both flamegraphs and benchmarksjs show a difference.


trxcllnt:
nice

PR #699, opened 2016-01-14
feat: toEsObservable and SymbolShim

gdi2290:


michaelbpaulson:
@gdi2290 
3 things:
1: there is some lint issues, please fix.
2: could you make the spacing in the symbol file be 4 spaces instead of 2!  Consistency.
3: How do we unit test this?

lib/SymbolShim.js
```diff
@@ -0,0 +1,58 @@
+var objectTypes = {
+    "boolean": false,
+    "function": true,
+    "object": true,
+    "number": false,
+    "string": false,
+    "undefined": false
+};
+
+var root = (objectTypes[typeof self] && self) || (objectTypes[typeof window] && window);
```
gdi2290: we probably need to tell the linter to ignore these

lib/SymbolShim.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+/*eslint-disable */
+var _root = (objectTypes[typeof self] && self) || (objectTypes[typeof window] && window);
+var freeGlobal = objectTypes[typeof global] && global;
+
+if (freeGlobal && (freeGlobal.global === freeGlobal || freeGlobal.window === freeGlobal)) {
+    _root = freeGlobal;
+}
+/*eslint-enable */
```
gdi2290: I'm suppressing eslint here since we need a hacky way to determine context

michaelbpaulson:
@gdi2290 to address point 3:  Unit tests, they are crashing!

michaelbpaulson:
@jhusain @gdi2290 At this point the unit tests in for properly aborting dataSources is in, you should be able to add the unit tests for rx 5 and get coverage up.  Once that is done, I'll merge this into master.

gdi2290:
awesome, I'll check it out tomorrow

michaelbpaulson:
Going to close this down as it now will be done in part of RxJS.  Then we just require in the module.

gdi2290:
Sounds good

PR #709, opened 2016-01-18
Add bower support for Falcor

JMBattista:
Add support for installing Falcor via bower by adding a bower.json file to the root and defining the main file.

I have registered the fork as falcor-fix in bower so you can check behavior via
```
bower install falcor-fix#0.1.16-alpha3
```

Once you add this bower.json future release tags will become available for install via bower.

Image of what gets installed
image

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,10 +1,9 @@
+
 {
   "name": "falcor",
-  "version": "1.4.2-build.4091+sha.fe0af2c",
+  "version": "0.1.15",
   "main": "./dist/falcor.browser.js",
   "ignore": [],
-  "dependencies": {},
-  "homepage": "http://netflix.github.io/falcor/",
-  "_source": "https://github.com/Netflix/falcor",
-  "_originalSource": "falcor"
+  "dependencies": {
+  }
 }
--- a/bower.json
+++ b/bower.json
@@ -3,7 +3,7 @@
   "name": "falcor",
   "version": "0.1.15",
   "main": "./dist/falcor.browser.js",
-  "ignore": [],
+  "ignore": ["build", "lib", "examples", "performance", "test"],
   "dependencies": {
   }
 }
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
 {
   "name": "falcor",
   "main": "./dist/falcor.browser.js",
-  "ignore": ["build", "lib", "examples", "performance", "test"],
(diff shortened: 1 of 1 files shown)


sdesai:
I don't see any reason not to include a bower package config for falcor.

I'm not too familiar with some of the real-world aspects of how to ideally package modules for bower, but a couple of thoughts, worth flushing out before we merge this in:

1. Would it be easier to explicitly include files, as opposed to ignore/blacklist files.
2. Do we need to give some thought to breaking up modules? Right now users get the falcor dist in it's entirety when you bower install falcor. Is this ideal? Or just a starting point, and we can consider more granular modules later if there's demand?
3. Do we need to roll up publishing to npm and bower as a combined publish task/step so semver versions stay in sync?

JMBattista:
Bower works by registering a name (in this case falcor) to a github repository (in this case https://github.com/Netflix/falcor/). It then detects the semver Git tags (see releases) and each tag becomes an installable version. When you perform the install (without version number) it finds the latest release tag and grabs the repository state for that tag and downloads it to the users system.

See: http://bower.io/docs/creating-packages/#register

1. The bower config only allows adding ignored files/folders. It follows the .gitignore syntax so there may be a way to 'exclude everything except'. [bower file spec](https://github.com/bower/spec/blob/master/json.md) 

2. I'm currently not sure when I'd want to use falcor.all vs falcor.browser, but the bower config only allows specifying one main file for each type. If you wanted to establish a more granular system, for example: falcor -> falcor.browser falcor-all -> falcor.all you could follow the example from the AngularJS team. They have a separate public repository, [bower-angular](https://github.com/angular/bower-angular) and have the bower registry pointed at that repo. In this way you can control the set of files that are installed for each command w/o any effect on your main repository (this also helps with 1 above). You would have to modify your CI job to push the dist to that repository and keep its tags up to date however.

3. Since Bower works off the tags in the git repo, so if those are currently insync w/ npm bower will be in-sync as well. If you want to pursue settings up individual bower-falcor repositories (re: 2 above) you would need to update the bower-falcor repository at the same time as publishing to npm to keep them in sync.

michaelbpaulson:
@sdesai if you are ok with merging this, feel free to hit the merge button.  I'll defer to you.

michaelbpaulson:
Then, of course, close #709 when you are finished here.

michaelbpaulson:
@sdesai is not around, so I'll just assume that he agrees with this PR.

Issue #678, opened 2015-12-30
Undefined response when expanding index range on model.get


khy:
When I do this:
```
model.get(['users', {from: 0, to: 10}, ['firstName', 'lastName']]).then( response => {...
```
`response.json.users` has three user objects (there are only three user objects in the database).

When I subsequently do this:
```
 model.get(['users', {from: 0, to: 20}, ['firstName', 'lastName']]).then( response => {...
```
`response.json.users` is undefined

On the second request, my falcor-express backend receives indexes 11 through 20 and returns an empty array which, as far as I can tell, is what it should do (since there are only three user objects).

My expectation is that on the second request, I would get back the same three user objects from the first request.

Is that expectation incorrect?

(let me know if this belongs in falcor-express)

trxcllnt:
@khy have you seen #590?

michaelbpaulson:
Its like it, but different.  Just has to do with `get`.  I can look at this soon.  Until then I'll be holidaying hard :).  It will be until early January (like the 5th).

khy:
Cool, have fun. No urgency, just experimenting.

michaelbpaulson:
I'll still get to this soon.  I am off on a non falcor tear for a bit.

michaelbpaulson:
@khy should be fixed here: https://github.com/Netflix/falcor/pull/714

PR #714, opened 2016-01-20
Fix materialized issue

michaelbpaulson:
Fixes: https://github.com/Netflix/falcor/issues/678


michaelbpaulson:
Turns out merging 0.x would be useless as the code base is becoming more and more different.

Issue #708, opened 2016-01-18
Is there a way to not include $__path key in the response payload


YesaDyuss:
When I am using falcor-htttp-datasource, the response payload contains `$__path` key everywhere. And I always have to filter it out. Is there a way to exclude it?

jhusain:
That key exists there to support the new deref function. It allows you to create a Model bound to an object anywhere in the message:

model.get("titlesList[0].name").then(v => {
  var listItem = model.deref(v.json.list[0]);
  return listItem.get("rating").then(p => log(p));
});

There's no way to hide it at the moment. I'm curious why you want to hide it. We considered adding in mode that would set enumerate to false, but I found it impacted performance.

YesaDyuss:
Is there an option to opt out of it. I just dont like all the filtering happenning in components

michaelbpaulson:
@YesaDyuss I am curious, why is this a problem?  I would imagine this is only a problem if you are doing the following type of init'ing

```javascript
var model = falcor.Model(...);

... later in code at a render function ...
var json = this.props.json;
return json.map(list => {
    return (
        <ListComponent json={list}></ListComponent>
    );
});
```

Then yes I could see this being a problem.  Is that what is happening?  if not, please elaborate why this is an issue.  

Second, this should have nothing to do with falcor-http-datasource.  This has everything to do with the falcor.Model.  The http data source simply will take a list of args and call an end-point via http with the remotable falcor operation.  Either the endpoint being called is returning a model payload (a no no, use a [falcor-router](https://www.github.com/Netflix/falcor-router) or this is the data from the Model's output.

Let me know!  But know that it will be a minimum 30% performance cost to use `defineProperty` to make `$__path` non-enumerable.  Hence the reason why it has never been an option.  But if this is truly a problem, then yes, we can consider adding one.

YesaDyuss:
Problem is roughly the same as you described: I pass `_.values(json.movies)` to `ListComponent`. And it is happening whenever I need to iterate over some collection. Now when upgraded to new falcor my components need to know about the existence of `$__path` and ignore them inside the component, or I need to filter them out when I am passing them to the component.

It is not a huge problem, but it is just annoying to change that in every place where I iterate over them and it loses some declarativeness, because web designers and new developers will be asking a question why are they using this dummy filter: `_.values(json.movies).filter(movie => movie.title).`.

To conclude I think $__path is related to falcor internals and it should not polute the output, that the user of the lib receives

[...]

PR #721, opened 2016-01-23
The helper function for getting the keys out of a falcor model response.

michaelbpaulson:


michaelbpaulson:
#708 Is partially addressed here.  This is a helper function that can return the list of keys for you instead of the user doing `Object.keys` you will do `falcor.keys`.  I know this is not the most ergonomic way of handling things, but it is the most unbreakable way of doing things right now.  We, that is the falcor team, are going to have a meeting about this next week and discuss the options.  What ever we come up with, this helper function will have to be available.

How does that sound @YesaDyuss ?

michaelbpaulson:
@YesaDyuss I am going to assume silence is agreeing and will be merging this in now and closing down #708.  We will come up with more stuff on top on this, but for now, every solution we see going forward will require this piece of code to be in the code base.

YesaDyuss:
@michaelbpaulson Apologies for not responding. Sounds great to me

PR #726, opened 2016-02-04
Some of the final changes for 1.x

michaelbpaulson:


michaelbpaulson:
Though curious as how documentation changes have caused a 0.04% decrease in code coverage, it is currently not something I am to worried about.

Issue #572, opened 2015-10-13
Expanding the falcor cache into plain JS objects


ekosz:
Hey all. I'm currently creating [a new library](https://github.com/ekosz/redux-falcor) to help connect falcor and [redux](https://github.com/rackt/redux). What I am trying to accomplish is allowing developers to pull data from the single redux store synchronously like normal, but having some of that data get populated by falcor.

You can see how I'm doing that [here](https://github.com/ekosz/redux-falcor/blob/7bbc617df9fd1dd256da1ad8f1e8391c1cc9d4b6/src/reducer.js). My current approach is a little naive. As results come back from falcor I just merge them into a master JS object that redux components can then pull from. This unfortunately loses much of the power that falcor provides like its expiration and LRU caching. 

What I would like to do instead would be something like:

``` js
falcor.onChange(() => store.dispatch(updateFalcor(falcor.expandCacheToJS())))
```

That has the advantage of always keeping the redux store update to date with the falcor cache, but it also means we're doubling (or more) the memory footprint as we're storing the cache data twice.

Another option I can think of is using ES6 [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) to create an object that hides the falcor implimentaion. But that would still require a way of getting the JS values out of the cache in a synchronous manner.

Do does anyone have any suggestions? I think falcor + redux is a powerful combination once the data sharing aspects are solved.

falcor-build:
We are actually exploring a very similar problem at the moment. Maybe we
could do a hang out or something. It would be interesting to share some
ideas.

JH

On Oct 13, 2015, at 11:09 AM, Eric Koslow <notifications@github.com> wrote:

Hey all. I'm currently creating a new library
<https://github.com/ekosz/redux-falcor> to help connect falcor and redux
<https://github.com/rackt/redux>. What I am trying to accomplish is allow
developers to pull data from the single redux store synchronously like
normal, but having some of that data instead get populated by falcor.

You can see how I'm doing that here
<https://github.com/ekosz/redux-falcor/blob/7bbc617df9fd1dd256da1ad8f1e8391c1cc9d4b6/src/reducer.js>.
My current approach is a little naive. As results come back from falcor I
just merge them into a master JS object that redux components can then pull
data from. This unfortunately loses much of the power that falcor provides
like expiration and its LRU caching.

What I would like to do instead would be something like:

falcor.onChange(() => store.dispatch(updateFalcor(falcor.expandCacheToJS())))

[...]

PR #730, opened 2016-02-09
Make Model#_clone return instance of current type

greim:
If I subclass `Model` like so:

``` js
class MyModel extends Model { ... }
var model = new MyModel().batch();
```

`batch()` nevertheless returns an instance of `Model`. This PR makes it so `batch()` and other cloning methods would return instances of `MyModel`.

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -425,7 +425,7 @@ Model.prototype._syncCheck = function syncCheck(name) {
 
 /* eslint-disable guard-for-in */
 Model.prototype._clone = function cloneModel(opts) {
-    var clone = new Model(this);
+    var clone = new this.constructor(this);
     for (var key in opts) {
         var value = opts[key];
         if (value === "delete") {


jhusain:
Just curious: why would you subclass model? Presumably there are many methods which return models which would also have to return the subclassed type. I would prefer to think of this class as sealed until we can think through all the implications. It would really help me if you could help me understand your use case.

trxcllnt:
@jhusain I've subclassed model to add an async invalidate, an inspect method, and overrides that all convert their path syntax arguments to arrays and return RxJS-5 Observables.

trxcllnt:
I also overrode `_clone` to return my subclass, so at least it's possible.

greim:
My use case was to have a model with a `getValueSync()` method which returns a value from the cache so that I can use the Falcor model directly as a store, ALA this thread: https://github.com/Netflix/falcor/issues/572

If you're mulling the best approach please don't feel pressured to merge this PR. I'm more keenly interested in the outcome of the above issue thread.

steveorsomethin:
Sounds like there's use cases for this, and I don't see any negatives. Gonna merge tomorrow. Any objections, @jhusain?

jhusain:
Primary concern is that I don't want to encourage subclassing to add a method to the Model. I'm reluctant do this because developers should treat Models like values, not objects. 

Subclassing the Model doesn't play well with the Model's compositional API. For example, what happens when you subclass a Promise? Does then() return MySubclassedPromise? Relying on overriding _clone to get this right depends on a private API. The compositional nature of the Model API suggests developers just pass Models to functions rather than add instance methods and try and override any method that returns a Model to preserve the subclass prototype. That's why I'm opposed to merging this PR.

trxcllnt:
@jhusain subclassing to override the core methods to return [Rx Observables/automatically parse path syntax etc.](https://github.com/graphistry/falcor/blob/528cb94fc343a4fe9dd00ce5370c2a03da4842c3/packages/falcor-model-rxjs/src/Model.js#L8) has been useful. Supporting this let me remove that stuff from the core, but still get the features in the apps where we can afford the extra cost. Just some feedback from using falcor in the wild, obviously do whatever in this project.

asyncanup:
A relevant discussion point here might be that Router does explicitly encourage subclassing (for performance gains): https://netflix.github.io/falcor/documentation/router.html#creating-a-router-class

jhusain:
I’m not sure that the Router API is relevant, because the Router and the Model are used very differently.  I think the central question here is this: do we want to make the Model safe to subclass?  This decision is not free. It would limit our ability to move away from a constructor to a factory-based API in the future, because such a move could impose large migration costs on developers which took advantage of an API we provided.

We should also consider what message this API sends to our users. API design is one of the most effective ways we have of communicating with our users and pushing them into the “pit of success”. As long as Models are sealed, we can ensure they are cheap to create, which in turn promotes the correct use of Models. Enabling subclassing allows developers to undermine this guarantee, which could cause them to be much more precious about their Models than we intend. There are also other questions to think about and answer. Do subclasses have access to Model internals (ie are any members protected)? If not, why not? 

It’s important to note that the real ask here does not appear to be able to subclass a model in the traditional sense (i.e. add additional state, write polymorphic implementations of methods). Instead it’s my perception that subclassing is simply a means by which to invoke stateless helper methods as though they were instance methods on the Model (please correct me if I’m wrong @greim). This is very reasonable given that this...

~~~js
value.do().this()
~~~

...is a lot more readable than this:


~~~js
this(do(value));
~~~

While I sympathize with this ask, it is a problem which all JavaScript frameworks suffer from equally. The root problem is a language problem: JavaScript privileges instance methods with a more ergonomic syntax. It’s my belief that this problem should be solved at the language level rather than the framework level (pipeline operator proposal https://github.com/tc39/proposal-pipeline-operator). Arguably this can already be accomplished relatively easily using ES2015 proxies.

Allowing Models to be subclassed in order to solve a JavaScript ergonomics problem opens a very wide door. How do we explain that we’re opening up subclassing, but don’t expect it to be used for any purpose but calling helper methods? There are a lot of questions to be answered here, and ultimately I think our efforts are best spent elsewhere.

jhusain:
@greim Wanted to add that if your main use case is to your invoke helper methods as if they're members of Model, this has been enabled with ES6 Proxies. You can create a Proxy for the model, intercept incoming methods, and dynamically implement them. While not as ergonomic as the pipeline operator, this should allow you to call helper methods you own as if they were instance methods on the Model. Remember to intercept calls to `deref`, `batch` to make sure the Model returned from those methods is also wrapped in your ES6 Proxy.

We got together and discussed this in more detail. The team collectively decided that we wanted to encourage developers to think of Models as cheap proxies, and consequently we want to discourage inheriting from Model. Under the circumstances we're moving away from constructors to factory functions in forthcoming versions.

asyncanup:
@greim 
After further discussion, we have decided to re-open this and merge it in.
The idea is not to encourage subclassing, but to still fix this pitfall that you fell into, and others may too.
Hope this helps!

Issue #728, opened 2016-02-08
Duplicate execution of call()


greim:
Here's an example:

```js
let count = 0;
const sourceModel = new Model({ cache: { a: 'b' } });
const source = sourceModel.asDataSource();
source.call = function() {
  count++;
  const fakeResp = sourceModel.get('a');
  return fakeResp;
};
const model = new Model({ source });
const resp = model.call('foo', ['baz'], [], []);
await resp;
await resp;
assert.strictEqual(count, 1);
```

...which gives:

```
AssertionError: 2 === 1
```

Internally, `await resp` calls `resp.then()`, which presumably is just an accessor, not an executor. I've had a report that this also happens with an actual HttpDataSource, not just the mocked-up one shown here. https://github.com/greim/ng-falcor/issues/6

(Falcor 0.1.16 running inside Node ~~0.4.x~~ 4.x. My test case is also being run through Babel.)

michaelbpaulson:
The `.then` method is technically "lazy" because its based off of observables.  I can do some research, but my guess is that, due to the lazy repeatable nature of obsversables, that the `.then()` execution multiple times causes multiple calls to the dataSource on call.  

Second, just for funsies, `call` will never be batched, its always sent the moment a subscribe / then is performed.

michaelbpaulson:
Let me take a look at the source codes and verify my assumptions.

michaelbpaulson:
Yes, its exactly what I thought it was doing.  I don't think this problem is solvable.  Perhaps I am not thinking clearly about this, but due to the lazy nature of falcor requests, `.then`s will kick off a wholly new promise each time.  If I am crazy, please re-open.

michaelbpaulson:
https://github.com/Netflix/falcor/blob/master/lib/response/ModelResponse.js#L99

greim:
I don't want to reopen if it's the intended behavior, but the duplicate execution definitely seems contrary to how thenables normally work. What about something like this?

```js
ModelResponse.prototype.then = function then() {
  if (!this._promise) {
    this._promise = new falcor.Promise( ... );
  }
  return this._promise;
};
```

That would preserve the laziness, while avoiding the duplication. I think the lazy behavior is fine.

michaelbpaulson:
@greim I have had a long day of travel.  You are completely right and now I am ashamed of my previous answer.  You should just make a PR and do exactly what you did above :)  Sorry, no sleep 30+ hours.

greim:
Haha no worries. Done!

PR #731, opened 2016-02-09
Don't re-execute a call on multiple then()s.

greim:
Fixes https://github.com/Netflix/falcor/issues/728

Tried to capture it with a test case. It failed before making the fix, now it passes. Thanks!

@michaelbpaulson


michaelbpaulson:
Nice test. +1

PR #733, opened 2016-02-12
Update webWorkerSource to work with the the newest falcor

### LINKED ISSUE #727: Web worker example doesn't work
Using https://github.com/Netflix/falcor/blob/master/examples/datasource/webWorkerSource.js as an example against latest master, I get the error:
`ERRORS: {"message":"An exception was thrown when making a request."}`

Which can be traced back to the lack of a falcor.Observable class.

Could the example be updated or some pointers on how to achieve a similar result?
/cc @michaelbpaulson

jontewks:
Updates webWorkerSource example:
* Use Rx instead of the previously included version of Rx attached to falcor
* Use the correct to JSONG function
* Add additional clarifying comments

This fixes #727

### DIFF
--- a/examples/datasource/webWorkerSource.js
+++ b/examples/datasource/webWorkerSource.js
@@ -1,8 +1,13 @@
- // In this example we demonstrate the communication between a model source and a server over a web worker
+// In this example we demonstrate the communication between a model source and a server over a web worker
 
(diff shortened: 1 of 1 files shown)


michaelbpaulson:
@jontewks I am going to assume that this works.  :)  if not, i'll make an angry pr.

PR #735, opened 2016-02-15
Adding Rx5 Support

jhusain:
Should now be possible to convert any Observable emitted from Falcor into an Rx5 Observable.


trxcllnt:
yes

PR #746, opened 2016-03-03
Lock bootstrap dependency version

### LINKED ISSUE #742: Documentation doesn't build
I've got a reproducible failure on the current head of the `gh-pages` branch (commit 50a5e500e0853fe3688203a6adbc7b4f4829cabf) using the following Dockerfile:

```
FROM node:4.3.1-onbuild
RUN apt-get update
RUN apt-get install -y rubygems ruby-dev
RUN gem install jekyll
RUN npm install -g gulp
CMD ["gulp", "serve"]
```

If I try to start this container up, I get this:

```
[07:14:29] Using gulpfile /usr/src/app/gulpfile.js
[07:14:29] Starting 'compile-less'...
[07:14:29] Starting 'js'...
[07:14:29] Finished 'js' after 656 ms
{ [Error: variable @dl-horizontal-breakpoint is undefined in file /usr/src/app/node_modules/bootstrap/less/type.less line no. 214]
  type: 'Name',
  filename: '/usr/src/app/node_modules/bootstrap/less/type.less',
  index: 4161,
  line: 214,
  callLine: NaN,
  callExtract: undefined,
  column: 21,
  extract:
   [ '',
     '  @media (min-width: @dl-horizontal-breakpoint) {',
     '    dt {' ],
  message: 'variable @dl-horizontal-breakpoint is undefined in file /usr/src/app/node_modules/bootstrap/less/type.less line no. 214',
  stack: undefined,
  lineNumber: 214,
  fileName: '/usr/src/app/node_modules/bootstrap/less/type.less',
  name: 'Error',
  showStack: false,
  showProperties: true,
  plugin: 'gulp-less',
  __safety: { toString: [Function: bound ] } }
```

A little bit of research turns up an [issue raised on Bootstrap](https://github.com/twbs/bootstrap/issues/13948).

Modifying the `package.json` file to restrict the `bootstrap` version to exactly `3.3.5` (instead of `^3.3.5`) makes everything build just fine. This seems to be a decent short-term fix for people wanting to make small changes.

The build failure, though, seems to indicate that the `variables.less` file (or `bootstrap-variables.less` file?) in the repo was made by copying the one out of Bootstrap version 3.3.5 and modifying it for Falcor's needs. This practice seems to be very much discouraged. The long-term fix, I think, would be to do what user `orrd` suggests on the Bootstrap issue

> The solution is that rather than customizing variables.less, what you're supposed to do is still let boostrap.less include the variables.less file, and then customize your Bootstrap variables in a separate file or somewhere else that is included after boostrap.less is included, so that your variables override just the ones you want to change, but you still let it also define all of the variables in the standard variables.less.

nicolasartman:
Fixes #742

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "homepage": "https://github.com/Netflix/falcor",
   "devDependencies": {
-    "bootstrap": "^3.3.5",
+    "bootstrap": "3.3.5",
     "browser-sync": "^2.8.0",
     "gulp": "^3.9.0",
     "gulp-autoprefixer": "^2.3.1",


nicolasartman:
@michaelbpaulson @jhusain or @sdesai mind merging this when you have a moment? Thanks!

PR #748, opened 2016-03-04
0.x change for removing boxValues from asDataSource().

michaelbpaulson:
toDataSource should use _jsonGraph() as output which properly does boxValues when needed.


steveorsomethin:
Double checked this looked good locally even though CI is angry.

PR #766, opened 2016-04-06
Avoid mutating the previous result object during progressive get.

lrowe:
Only for JSON responses, not JSONGraph. Structural sharing of unchanged subobjects.


jhusain:
Code looks good to me. Will need to follow up on the Travis build failure.

lrowe:
Logic moved into get. I've added error handling for results.criticalError in setRequestCycle, is returning null in this case the right thing to do?

PR #771, opened 2016-04-08
WIP Invalid deref

lrowe:
Currently TypeErrors are thrown from property access of undefined for invalidated deref path get and set. For get we should return a criticalError, but I'm not sure what we should do for set.

### DIFF
--- a/lib/get/get.js
+++ b/lib/get/get.js
@@ -32,7 +32,7 @@ module.exports = function get(walk, isJSONG) {
 
             // If there was a short, then we 'throw an error' to the outside
             // calling function which will onError the observer.
-            if (currentCachePosition.$type) {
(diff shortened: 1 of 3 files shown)


trxcllnt:
@lrowe fwiw I have updated set impls in my [secret-ish experimental branch](https://github.com/trxcllnt/falcor/blob/the-refsets-work/lib/set/setPathValues.js) that are more closely aligned with the throw-ing behavior of get.

steveorsomethin:
@lrowe is this still relevant? Is there an issue referencing this? Looks like it might be stale with recent changes.

lrowe:
We should run the tests added in this branch on current master to see if this is still an issue.

asyncanup:
@lrowe just tested. the tests don't succeed on latest `master`

asyncanup:
Closing in favor of #940

PR #777, opened 2016-04-19
Eslint update

lrowe:
Update to eslint 2. The first four commits fix consistent-return linting errors.


.eslintrc
```diff
@@ -43,8 +43,8 @@
         "consistent-return": [ 2 ],
         "curly": [ 2 ],
         "default-case": [ 2 ],
-        "dot-notation": [ [ 2 ], { "allowKeywords": true } ],
-        "eqeqeq": [ 2 ],
+        "dot-notation": [ 2, { "allowKeywords": true } ],
+        "eqeqeq": [ 2, "smart" ],
```
yunong: Why not just use === across the board?

lrowe: Just the relatively large number of errors it throws up (26.)

yunong: Is it worth while fixing? Seems like an easy fix for literals.

lrowe: I'd like to avoid the work of sorting out which of the various `!= null` comparisons need to also take account of undefined.

PR #778, opened 2016-04-21
Avoid require graph cycles by moving Promise shimming to build step.

lrowe:
Picking a Promise shim is now a build time choice, allowing the ~1KB page weight of the 'promise' package to be avoided for those who don't require it.


DonutEspresso:
Thanks for quick response @lrowe! 
FYI @memoryhole

lrowe:
If  we want to keep Node 0.10 support then we can shim in Promise in the test setup.

sdesai:
We should add a migration writeup for this (to MIGRATION.md) for people building from src (or node who were expecting a shim), who want a Promise shim. Basically something which covers what the change means for end users and our support matrix. And visa versa, what it means in terms of a k-weight win.

lrowe:
Testing in the browser (for posterity):
```html
<html>
<head>
<script>
    delete window.Promise; // check with and without Promise
</script>
<script charset='utf8' src="falcor.browser.js"></script>
<script>
    var model = new falcor.Model();
    model.get().then(() => console.log('foo'));
</script>
</head>
<body>
</body>
</html>
```

From the JS console check that Promise has not leaked into global scope.

PR #780, opened 2016-04-23
chore(package.json): correct package version

gdi2290:


sdesai:
I was avoiding bumping master up to 0.1.17, since that release is cut off the 0.x branch (where hopefully it should reflect 0.1.17).

Open to suggestions on what the master package.json should reflect on 'next' release branches, based on other projects. Since master has back-compat breakage (and we wanted to avoid continuing to use 0.x for any changes we publish from it), figuring whatever we release off master next will be 1.x, so we could have it be 1.0.0-rc1 or something along those lines.

steveorsomethin:
Gonna close since we're post 1.x now.

gdi2290:
np

PR #786, opened 2016-04-30
Fixes setPathValues when setting with null as last key and hard references.

trxcllnt:
This PR fixes the issue that set would improperly short-circuit if setting beyond a reference that points to a value if that reference was already hard-linked to the value.

### DIFF
--- a/test/set/pathValues/atom.spec.js
+++ b/test/set/pathValues/atom.spec.js
@@ -167,6 +167,25 @@ describe("an atom", function() {
         }));
     });
 
+    it("through a reference that lands on an atom with a null last key", function() {
+
+        var cache = {};
+        var version = 0;
+        setPathValues(
+            getModel({ cache: cache, version: version++ }), [
+                $pathValue("foo.value", $ref("bar.value")),
+                $pathValue("bar.value", $atom("bar-value")),
+                $pathValue(["foo", "value", null], $atom("foo-value-1")),
+                $pathValue(["foo", "value", null], $atom("foo-value-2"))
+            ]
+        );
+
+        expect(strip(cache)).to.deep.equal(strip({
+            foo: { value: $ref("bar.value") },
+            bar: { value: $atom("foo-value-2") }
+        }));
+    });
+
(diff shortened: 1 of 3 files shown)


trxcllnt:
@sdesai I'm not sure why travis is failing the integration test for this PR, as it finishes cleanly on my machine. I know @jhusain is OOO at the moment, and you're probably swamped with falcor support, but is there anything more I can do to help get this fix in to master?

sdesai:
Hey Paul, I'll take a look at it, and merge it in if it's all good.

trxcllnt:
@sdesai Thanks, I appreciate it.

test/set/pathValues/atom.spec.js
```diff
@@ -167,6 +167,25 @@ describe("an atom", function() {
...
 
+    it("through a reference that lands on an atom with a null last key", function() {
+
+        var cache = {};
+        var version = 0;
+        setPathValues(
+            getModel({ cache: cache, version: version++ }), [
+                $pathValue("foo.value", $ref("bar.value")),
+                $pathValue("bar.value", $atom("bar-value")),
+                $pathValue(["foo", "value", null], $atom("foo-value-1")),
```
sdesai: Just getting around to looking at this. Sorry about the delay.

What's the motivation in setting `["foo", "value", null]` twice as part of the test?

Trying to make sure I understand the change. I believe I understand the second set to foo-value-2, and the subsequent assertion (to make sure the reference is maintained, and the atom is the thing which actually gets set/overwritten), but I didn't follow what the first set to foo-value-1 is trying to validate.

trxcllnt: @sdesai the first set to `["foo", "value", null]` is ensuring the hard-link is assigned from `["foo", "value"]` to `["bar", "value"]`. `setPathValues` is writing the first two PathValues directly, but the hard-link is only set up once the reference (`foo.value`) is crossed during a get or set. I wrote two distinct values ('foo-value-1' and 'foo-value-2') to validate that the second value was written by crossing the hard-link.

PR #790, opened 2016-05-25
Fix transpilation error to support React Native

### LINKED ISSUE #603: Make falcor work with react-native
Due to `asap` dependency, building a `react-native` project will fail. 
`relay` with `react-native` has the same issue, described in details here:

https://github.com/facebook/react-native/issues/3463

clarle:
Recent versions of Babel will throw a TransformError in certain runtimes if any property on `Symbol` is assigned a value using dot notation, which includes React Native.  This makes sense, since you probably don't want to be assigning values on top of existing symbols anyway in environments that already support them.

screen shot 2016-05-25 at 12 53 43 am

This retains the same polyfilling behavior for platforms without `Symbol.for` and `Symbol.observable`, but lets Babel properly transpile.  

With the change to Falcor to support Promise shimming merged in, this is the last remaining issue needed to support React Native in 1.x. Fixes #603.

@jhusain @sdesai @ktrott

### DIFF
--- a/lib/SymbolShim.js
+++ b/lib/SymbolShim.js
@@ -28,13 +28,15 @@ function ensureSymbol(root) {
 }
 
 function ensureObservable(Symbol) {
+    /* eslint-disable dot-notation */
     if (!Symbol.observable) {
         if (typeof Symbol.for === "function") {
-            Symbol.observable = Symbol.for("observable");
+            Symbol["observable"] = Symbol.for("observable");
         } else {
-            Symbol.observable = "@@observable";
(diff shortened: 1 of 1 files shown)


sdesai:
Looks fine to me if it gets us by Babel. All the native symbols (iterator, match, etc.) seem to be "writable:false" natively.

It seems to be heavy handed with `Symbol.for` but that's fine [ I imagine that's just a 'less transpiler implementation complexity' call ].

Will give Jafar a little to comment, otherwise will merge later today

PR #794, opened 2016-06-15
Fix documentation for Authorization

stevelacy:
- adds `.prototype` to fix Object instantiation. https://github.com/Netflix/falcor/compare/gh-pages...stevelacy:gh-pages?expand=1#diff-baef559ba7d4197065edabd0af8baae0R348
- removes trailing whitespace


steveorsomethin:
Thanks!

PR #815, opened 2016-10-14
Prevent max call stack size error; Fixes #801

### LINKED ISSUE #801: Get Cache getCache.js:45 Maximum call stack size exceeded
Not entirely sure why its happening but _copyCache, gets to the bottom and then wants to copy every character in a string, which the output doesn't have keys for obviously and so this function falls into an infinite loop and returns max call stack.

I'm adding
```
if(typeof node !== 'object') return;
```  
to the top of _copyCache to get around it. 

not sure if this is like an invalid graph structure problem on my part, happy to follow up with details if requested.

kevnk:
only copy objects. See https://github.com/Netflix/falcor/issues/801 for more details

### DIFF
--- a/lib/get/getCache.js
+++ b/lib/get/getCache.js
@@ -33,7 +33,7 @@ function _copyCache(node, out, fromKey) {
     // copy and return
 
     // only copy objects
-    if (node && typeof node !== "object") {
+    if (!node || typeof node !== "object") {
         return;
     }
 
--- a/lib/get/getCache.js
+++ b/lib/get/getCache.js
@@ -33,10 +33,10 @@ function _copyCache(node, out, fromKey) {
     // copy and return
 
     // only copy objects
-    if (typeof node !== 'object') {
+    if (typeof node !== "object") {
         return;
     }
-    
+
     Object.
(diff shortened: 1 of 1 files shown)


lrowe:
Are you able to reproduce this error? Any chance you could add a test case?

steveorsomethin:
@lrowe, much as I'd like a test case, in reading the code this looks like a legit bug. I'd like to merge it regardless, as the PR is quite stale. Objection?

asyncanup:
LGTM.
@jhusain @lrowe your thoughts?

sdesai:
I'd recommend seeing if we can set up a test case which hits the failure case (as in defining a valid cache state which would cause the issue), if we can't get one. That way if the code gets moved around in the future we don't regress.

lrowe:
I don't think this fix is complete as current since `Object.keys(null)` throws a TypeError.

asyncanup:
Website team saw this happen as well when they had an extra custom field on their cache nodes (legacy reasons). i think this PR is not needed, instead simply avoid setting any custom fields on cache nodes.

PR #821, opened 2016-11-24
Fix case where error is returned with wrong path

lrowe:
Looking at get.js I suspect there may be similar problems with derefInfo and optimizedPath. Alternatively we could avoid mutating requestedPath/branchInfo/optimizedPath in walkPath.js.


lrowe:
I've switched approach to preserve the performance optimization of reusing arrays during walk. We now consistently increment depth when appending null.

The reusing of branchInfo and optimizedPath should be safe.

jhusain:
LGTM

PR #841, opened 2017-04-13
Remove unicode characters from source code

jhusain:
Falcor used to use the smiley face unicode character for private members which we placed on objects we returned to users and for bookkeeping information on cache objects. The intention was to avoid conflicting with user input, but the unicode characters wreaked havoc on a variety of tools.

With this change we've decided to reserve the "$" keyspace. When inserting user input which may start with '$' into a path, the input should be encoded. The following code...

```js
async function run() {
  await {json} = await model.get(["search", input, {from:0,to:10}, "name"]);
  console.log(json.search[input][0].name);
}
```

...must be rewritten like this...

```js
async function run() {
  let escapedInput = escape(input);
  await {json} = await model.get(["search", escapedInput, {from:0,to:10}, "name"]);
  console.log(json.search[escapedInput][0].name);
}
```


jhusain:
Submitting a cleaner PR with no prettier-related formatting changes.

PR #846, opened 2017-05-17
No longer creating Path-bound errors for DataSource errors

jhusain:
**BREAKING CHANGE**: Model no longer creates path-bound errors when DataSource onErrors

In the current Master HEAD if a DataSource's Observable onError's, the Falcor Model will create a path-bound error for every requested path, and insert a JSON Graph error object at each requested path in the Model cache. **The problem with this approach to error handling is that it can create the false impression that objects exist which do not exist.**

As an example, the code below attempts to retrieve two paths from a DataSource which always onErrors.

```js
import { Observable } from "rx"
const FailingDataSource = { 
  get() { return Observable.error({ message: "FAIL" }); }
};

async function test() {
  const model = new falcor.Model({ source: FailingDataSource }):
  try {
    await model.get(["videos", [123,456], "name"]);
  }
  catch(errors) {
    console.log(errors);
  }
}

test();
```

The Falcor Model creates an error for each requested path. As a result, the following array is onError'ed and printed to the console:
```
[
  { path: ["videos", 123, "name"], value: { message: "FAIL" } },
  { path: ["videos", 456, "name"], value: { message: "FAIL" } }
]
```

Furthermore the Model inserts JSON Graph error objects into the cache at each requested path.  This is the state of the Model cache after the code above executes:

```
{
  videos: {
    123: {
      name: { $type: "error", value: { message: "FAIL" } }
    },
    456: {
      name: { $type: "error", value: { message: "FAIL" } }
    }
  }
}
```
Note that placing a JSON Graph error at ["videos", 123, "name"] and ["videos", 456, "name"] respectively creates the impression that **both ["videos", 123] and ["videos", 456] exist**. This is __not__ necessarily true, bearing in mind that a DataSource error can be caused by things as innocuous as a timeout.

This PR halts the practice of creating JSON Graph errors for DataSource errors sent via a DataSource notification. The new behavior is that Model's will simply forward errors sent from DataSources as Promise rejections or Observable onError notifications. After this PR the code example above will produce the following console output:

```
// logged rejection from Model Promise
{ message: "FAIL" }
```

Furthermore no further data will be written into the Falcor cache when a Model recieves a DataSource error.

This is a **breaking change**, because the shape of error objects that come out of ModelResponses will change. Instead of the Model onError'ing an Array of errors wrapped in path-bound values, the Model will simply onError the error that comes out of the DataSource.

### DIFF
--- a/test/falcor/error/index.js
+++ b/test/falcor/error/index.js
@@ -18,19 +18,16 @@ describe("Error", function() {
         toObservable(model.
             get(["test", {to: 5}, "summary"])).
             doAction(noOp, function(err) {
-                expect(err.length).to.equal(6);
                 // not in boxValue mode
                 var expected = {
-                    path: [],
+                    $type: "error",
                     value: {
                         status: 503,
                         message: "Timeout"
                     }
                 };
-                err.forEach(function(e, i) {
-                    expected.path = ["test", i, "summary"];
-                    expect(e).to.deep.equals(expected);
-                });
+
+                expect(err).to.deep.equals(expected);
             }).
             subscribe(function() {
                 done('Should not onNext');
@@ -66,8 +63,8 @@ describe("Error", function() {
             get(["test", {to: 5}, "summary"])).
             doAction(onNext, function(err) {
 
-                // Ensure onNext is called correctly
-                expect(onNext.calledOnce, 'onNext called').to.be.ok;
+                // Ensure onNext is not called
+                expect(onNext.callCount, 'onNext called').to.equal(0);
                 expect(clean(onNext.getCall(0).args[0]), 'json from onNext').to.deep.equals({
                     json: {
                         test: {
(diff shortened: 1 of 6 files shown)


jhusain:
Looks like the CI check is just failing because we recently had to revoke OAuth tokens. Won't block on this.

test/falcor/error/index.js
```diff
@@ -18,19 +18,16 @@ describe("Error", function() {
         toObservable(model.
             get(["test", {to: 5}, "summary"])).
             doAction(noOp, function(err) {
-                expect(err.length).to.equal(6);
                 // not in boxValue mode
                 var expected = {
-                    path: [],
+                    $type: "error",
```
lrowe: These tests using ErrorDataSource are a little confusing, but looks like the `{$type: "error"...}` boxing is coming from the data source not the Model here.

PR #848, opened 2017-05-18
Removed oAuth token, using ssh instead

falcor-build:


steveorsomethin:
@sdesai @jhusain @lrowe what's the context on this one? Why wasn't it merged?

sdesai:
My recollection is that this made it to the code base already (maybe through some other PR? Directly?). It's worth taking a look at the yml and deploypages.sh impl on master and 0.x. If it hasn't it should be merged.

steveorsomethin:
Thanks, looks like it wasn't merged. Merging.

PR #850, opened 2017-06-21
Fix misspelling of JavaScript and change algorithm -> method

micahr:
In an effort to make the docs slightly more accessible, I've changed algorithm -> method where it is obviously referencing a method. Also fixed a misspelling of JavaScript


steveorsomethin:
@jhusain in context here though, it's only changing the terminology for .toString(), which subjectively to my eyes is an improvement. Framed that way, any reason not to merge this?

asyncanup:
i've updated the PR with latest from gh-pages.
i do agree with @steveorsomethin and @micahr that using `method` is better for `toString()`.

@jhusain @lrowe thoughts?

sdesai:
Yeah, merging.

PR #854, opened 2017-07-24
Omit references

steveorsomethin:
Previously, model.get(['a', 'b']) where b is a reference would result in references in the json output. This change instead results in emitting nothing where a reference would previously exist. JSONG remains unchanged.

### DIFF
--- a/lib/get/onValue.js
+++ b/lib/get/onValue.js
@@ -34,6 +34,11 @@ module.exports = function onValue(model, node, seed, depth, outerResults,
         valueNode = clone(node);
     }
 
+    // We don't want to emit references in json output
+    else if (!isJSONG && node.$type === $ref) {
+        valueNode = undefined;
+    }
+
     // JSONG always clones the node.
     else if (node.$type === $ref || node.$type === $error) {
         if (isJSONG) {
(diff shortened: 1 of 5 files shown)


lib/get/onValue.js
```diff
@@ -34,6 +34,11 @@ module.exports = function onValue(model, node, seed, depth, outerResults,
         valueNode = clone(node);
     }
 
+    // We don't want to emit references in json output
+    else if (!isJSONG && node.$type === $ref) {
+        valueNode = undefined;
```
lrowe: Do we want to emit an empty branch node here instead? Though perhaps not until we start emitting branch nodes we know to be present.

steveorsomethin: That was the original idea (and is simple to implement) but after a discussion with @jhusain yesterday the goal is to omit keys that are missing, atom of undefined, or ref.

ardeois:
@steveorsomethin is there any reason why `model.get` references are not emitted anymore?
I'm wondering what should the best practice be?
I've also added a question on stack overflow in case you want to answer https://stackoverflow.com/questions/54101216/falcor-get-no-longer-emits-references-as-leaf-values

Thanks !

PR #855, opened 2017-07-25
Always onnext

steveorsomethin:
Currently, a get() for paths that are missing in the cache behaves differently than a get() for atoms of undefined. This PR makes both conditions produce identical results, and ensures onNext() is always called on the model result.

### DIFF
--- a/lib/response/get/checkCacheAndReport.js
+++ b/lib/response/get/checkCacheAndReport.js
@@ -47,7 +47,7 @@ module.exports = function checkCacheAndReport(model, requestedPaths, observer,
     //
     // 2.  The request if finished and the json key off
     // the valueNode has a value.
-    if (progressive || ((progressive && results.hasValues || !progressive) && completed)) {
+    if (progressive || ((progressive && results.hasValues || !progressive) && completed && valueNode !== undefined)) {
         try {
             observer.onNext(valueNode);
         } catch (e) {
--- a/lib/response/get/checkCacheAndReport.js
+++ b/lib/response/get/checkCacheAndReport.js
@@ -17,9 +17,10 @@ var getWithPathsAsPathMap = gets.getWithPathsAsPathMap;
  * @private
  */
 module.exports = function checkCacheAndReport(model, requestedPaths, observer,
-                                              progressive, isJSONG, seed,
+                                              progressive, isJSONG, prevResults,
                                               errors) {
 
+    var seed = prevResults.values;
     // checks the cache for the data.
     var results;
     if (isJSONG) {
--- a/lib/response/get/checkCacheAndReport.js
+++ b/lib/response/get/checkCacheAndReport.js
@@ -54,7 +54,7 @@ module.exports = function checkCacheAndReport(model, requestedPaths, observer,
     //
     // 2.  The request if finished and the json key off
     // the valueNode has a value.
-    if (progressive || completed) {
+    if (progressive || ((results.hasValues || (errors.length === 0 && !results.criticalError)) && completed)) {
         try {
             observer.onNext(valueNode);
         } catch (e) {
--- a/lib/response/get/checkCacheAndReport.js
+++ b/lib/response/get/checkCacheAndReport.js
@@ -39,11 +39,10 @@ module.exports = function checkCacheAndReport(model, requestedPaths, observer,
     // We are done when there are no missing paths or the model does not
     // have a dataSource to continue on fetching from.
     var valueNode = results.values[0];
-    var hasValues = results.hasValue;
+
     var completed = !results.requestedMissingPaths ||
                     !results.requestedMissingPaths.length ||
                     !model._source;
-    var hasValueOverall = Boolean(valueNode.json || valueNode.jsonGraph);
 
     // Copy the errors into the total errors array.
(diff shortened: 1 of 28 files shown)


test/falcor/get/get.dataSource-and-cache.spec.js
```diff
@@ -582,7 +588,7 @@ describe('DataSource and Partial Cache', function() {
                         done();
                     },
                     function() {
-                        expect(onNextSpy.callCount).to.equal(0);
+                        expect(onNextSpy.callCount).to.equal(1);
```
jhusain: Can we confirm the contents of the body is an empty object?

lrowe:
Testing locally I see that we're not onNexting when we request no paths:
```
$ node
> var falcor = require("./lib")
undefined
> var model = new falcor.Model({cache: {foo: 1}})
undefined
> var sub = model.get().subscribe(console.log)
undefined
```

steveorsomethin:
Oops, good catch. Doesn't seem like someone would _actually_ want to request nothing, but I see a consistency argument. Will fix.

steveorsomethin:
Let me dig in a little more before you review further. The empty get thing exposed some additional problems.

steveorsomethin:
Ok, proceed.

lrowe:
Another minor weirdness:
```
$ node
> var falcor = require("./lib")
undefined
> var model = new falcor.Model({cache: {foo: 1}})
undefined
> var s = model.asDataSource().get(['a']).subscribe(console.log)
{ jsonGraph: { a: { '$type': 'atom' } }, paths: [ [ 'a' ] ] }
undefined
> var s = model.asDataSource().get([]).subscribe(console.log)
{ jsonGraph: {} }
undefined
```
I would expect the second JSONGraph response to include an empty paths array.

steveorsomethin:
Thanks, fixed + test.

test/falcor/set/set.dataSource-and-cache.spec.js
```diff
@@ -159,6 +159,64 @@ describe('DataSource and Cache', function() {
...
+                            genreList: {
+                                0: {
+                                    0: {
+                                        summary: 1337
+                                    }
+                                }
+                            }
+                        }
+                    });
+                }).
```
jhusain: Please assert against the value of the second onNext.

test/get-core/values.spec.js
```diff
@@ -187,10 +187,57 @@ describe('Values', function() {
...
+                videos: {
+                    123: {
+                        name: atom()
+                    }
+                }
+            }
+        });
+    });
+
     it('should have no output for empty paths.', function() {
```
jhusain: This test is probably not named appropriately given we are outputting data for empty paths.

test/get-core/values.spec.js
```diff
@@ -187,10 +187,57 @@ describe('Values', function() {
...
+                            
+                        }
+                    }
+                }
+            },
+            cache: cacheGenerator(0, 1)
+        });
+    });
+
+    it('should have no output for empty get.', function() {
```
jhusain: Test not named appropriately given we're outputting data.

test/integration/call.spec.js
```diff
@@ -113,7 +113,7 @@ describe('call', function() {
             call("genreList[0].titles.push", args)).
             doAction(onNext, noOp, noOp).
             subscribe(noOp, done, function() {
-                expect(onNext.callCount).to.equal(0);
+                expect(onNext.callCount).to.equal(1);
```
jhusain: Can we assert on onNext contents?

lib/response/get/checkCacheAndReport.js
```diff
@@ -17,33 +17,36 @@ var getWithPathsAsPathMap = gets.getWithPathsAsPathMap;
  * @private
  */
 module.exports = function checkCacheAndReport(model, requestedPaths, observer,
```
jhusain: Can you document the fields in seed?

lrowe:
## Re: should not onNext when only receiving errors / should onNext when receiving errors and missing paths (or values)

Are we being consistent between progressive and non-progressive mode here?

### Without treatErrorsAsValues

PBE's effectively become critical errors since view code does not want to account for them.

#### Non-progressive mode without treatErrorsAsValues

We only want to render the final view.

I don't see the purpose in onNext'ing any value when it will be followed by an onError. For this case I think developers usually want behaviour more like a Promise / Rx Single / async Result, so either onNext exactly once or onError but never both.

#### Progressive mode without treatErrorsAsValues

We want the opportunity to display a partial view before rendering the final view. If there are any errors then we render an error view instead.

What's the expected behaviour in progressive mode when:

1. No errors in cache and a mix of values and errors are returned by the datasource

    I would expect an immediate (possibly empty) onNext followed by an onError when the datasource returns even if there are values.

2. All errors in cache, no unknown values

    No need to make a datasource request, onError immediately.

3. Some errors in cache, some unknown values

    Do we even need to make a datasource request here? If we're going to render an error view we may as well display it immediately.

### With treatErrorsAsValues / boxedValues

In this case PBEs in the cache are treated as values so do not trigger an error. Only critical errors would result in same behaviour as above.

PR #868, opened 2017-09-03
Move rx to devDependencies

ratson:
As it is removed in v1.0.0 and only being used in test.


steveorsomethin:
Strange, thought this was already done. Thanks for the catch.

PR #874, opened 2017-09-08
Include missing paths in MaxRetryExceededError instances

asyncanup:
this will help developers understand data source and service failures better

### DIFF
--- a/lib/errors/MaxRetryExceededError.js
+++ b/lib/errors/MaxRetryExceededError.js
@@ -5,9 +5,9 @@ var NAME = "MaxRetryExceededError";
  *
  * @private
  */
-function MaxRetryExceededError(missingPaths) {
+function MaxRetryExceededError(missingOptimizedPaths) {
     this.message = "The allowed number of retries have been exceeded.";
-    this.missingPaths = missingPaths || [];
+    this.missingOptimizedPaths = missingOptimizedPaths || [];
     this.stack = (new Error()).stack;
 }
 
--- a/lib/errors/MaxRetryExceededError.js
+++ b/lib/errors/MaxRetryExceededError.js
@@ -5,8 +5,9 @@ var NAME = "MaxRetryExceededError";
  *
  * @private
  */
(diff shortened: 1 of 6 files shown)


lib/errors/MaxRetryExceededError.js
```diff
@@ -5,8 +5,9 @@ var NAME = "MaxRetryExceededError";
  *
  * @private
  */
-function MaxRetryExceededError() {
+function MaxRetryExceededError(missingPaths) {
     this.message = "The allowed number of retries have been exceeded.";
+    this.missingPaths = missingPaths || [];
     this.stack = (new Error()).stack;
```
DonutEspresso: Since we're in the client this may not be possible, but is there something like `captureStackTrace` available across different browsers?

lrowe: To remain cross platform I think `(new Error()).stack` is the best that can be done here.

lib/response/set/setRequestCycle.js
```diff
@@ -12,17 +12,18 @@ var MaxRetryExceededError = require("./../../errors/MaxRetryExceededError");
  */
 module.exports = function setRequestCycle(model, observer, groups,
                                           isJSONGraph, isProgressive, count) {
+    var requestedAndOptimizedPaths = setGroupsIntoCache(model, groups);
+    var optimizedPaths = requestedAndOptimizedPaths.optimizedPaths;
+    var requestedPaths = requestedAndOptimizedPaths.requestedPaths;
```
steveorsomethin: Something to consider which could be a little scary: we're going from conditionally setting groups into cache, to always setting. It might be that the shape of what is set makes this inconsequential, but could be worth a close look before you merge.

asyncanup: Talked it out with Jafar. It seems logical to add any potential results returned from server into cache, even if it was on the last retried request.
Good catch!

jhusain:
LGTM

PR #882, opened 2017-09-16
1.0.0

steveorsomethin:
If we're all good with this, I'll push the button. Final concerns?


jhusain:
LGTM

sdesai:
Fine with me. Lets just loop back around with the folks who have been trying it, to see if they have anything pop up, and may be we can publish it Monday?

PR #885, opened 2017-09-20
Runnable code samples!

asyncanup:
and some minor language changes.

Some notes:
- uses RunKit to make code samples runnable
- samples run Node.js based falcor on RunKit Node servers
- also uses a simple now.sh service (https://falcor-server-sample-nvutywuvbl.now.sh/_src) to make `HttpDataSource` samples work


documentation/model.md
```diff
@@ -1438,32 +1479,37 @@ var model = new falcor.Model({cache: {
...
-            console.error(JSON.stringify(pathValue));
-        });
+var response;
+try {
+    response = await model.
+        treatErrorsAsValues().
+        get('titlesById[44].name');
+} catch (error) {
+    console.log('error was thrown');
+    error
```
jhusain: Can we make sure to add return statements if the intention is to add data to the console? At the moment this looks like a noop or a bug?

documentation/model.md
```diff
@@ -1438,32 +1479,37 @@ var model = new falcor.Model({cache: {
...
+} finally {
+    console.log('error treated like a value:');
+    response;
```
jhusain: Please add return statement if possible

documentation/model.md
asyncanup: done

documentation/model.md
asyncanup: done

PR #886, opened 2017-09-25
Update README.md

steveorsomethin:


README.md
```diff
@@ -4,13 +4,17 @@
 
 # Falcor [![Build Status](https://travis-ci.org/Netflix/falcor.svg)](https://travis-ci.org/Netflix/falcor) [![Coverage Status](https://coveralls.io/repos/Netflix/falcor/badge.svg?branch=master&service=github)](https://coveralls.io/github/Netflix/falcor?branch=master) [![bitHound Score](https://www.bithound.io/github/Netflix/falcor/badges/score.svg)](https://www.bithound.io/github/Netflix/falcor)
 
-## Release Candidates
+## 1.0.0 has been released 🎉
```
sdesai: I would take the 'temporal' aspect out of the title, and test below, and just phrase it for the steady state. That is, "1.0.0 is the current Falcor release" or something along those lines. Would change "Users of the technical preview" to "0.x Users" or "0.1.17 Users".

Don't need to change anything about the 2.0.0 paragraph (originally we had discussed if we want to phrase it as a "fast-follow"). I think as soon as we've had a chance to validate it with internal users, we'd probably put it out there.

Feel free to merge after the 1.0.0 temporal changes.

PR #897, opened 2017-10-06
partial request deduping

### LINKED ISSUE #779: Path Deduping only works on subsets, not intersecting sets of paths
Currently path deduping only works if a set of requested paths is a _subset_ of the paths in an outstanding request. However we'd like deduping to work in the event that a set of requested paths intersects with those in an outstanding request, with only the difference being requested from the server.

For example, a request for ["genrelists", {from:0,to:5}, "name"] will be deduped against an outstanding request for ["genrelists", {from:0,to:10}, "name"]. However the request for ["genrelists", {from:0,to:5}, "name"] will _not_ be partially deduped against an outstanding request for ["genrelists", {from:1,to:4}, "name"], with only the difference being requested from the server (["genrelists", [0,5], "name"]).

When a batch of outgoing requests is being accumulated, it is modeled as an Array of Array of Paths. In other words two separate requests for ["genreLists", 0, "name"] and ["genreLists", 1, "name"] will be modeled like this in a batch:

``` js
var outgoingBatch = 
  [
    [ ["genreLists", 0, "name"] ], // first request
    [ ["genreLists", 1, "name"] ], // second request
  ];
```

When a batch is sent out, it is converted into a path map, making it efficient to check for intersections with other sets of paths:

``` js
var outstandingRequestPathMap =
  {
    genreLists: {
      0: { name: true }
      1: { name: true }
    }
  }
```

Each outstanding request has a count indicating how many outstanding observers are awaiting the response. If all observer's unsubscribe, the count drops to 0 and the request is cancelled.

Currently the paths in each new request is matched against each outstanding request's PathMap. If the set of requested paths is a _subset_ of an outstanding batch's set of paths, it is deduped and the number of observers listening for the batch is incremented. However if a set of paths merely intersects with those of the outstanding request, the algorithm gives up and creates a new batch for the paths.

The fix is to alter the algorithm to create a new batch for just the _difference_ between the requested paths and the sets of paths that have already been sent to the server. Then it will be necessary to ensure that the request is only processed after all outstanding request dependencies (ex. ["genreLists", {from: 1, to: 4}, "name"] and ["genreLists", [0,5], "name"]) hey have been received from the server and merged into the cache. @trxcllnt Do you have anything to add to this?

@ktrott @steveorsomethin @sdesai @lrowe

asyncanup:
update:
pull request is ready for review and merging! all edge cases covered.

update:
pull request is up for review now. all edge case bugs fixed. adding some tests now.

dedupes 2 major cases: paths followed via refs in cache, and paths with ranges in the end
pull request is still in progress, but still want to get opinion on the approach
work remaining: dedupe paths with more properties after the range, and not followed via refs in client cache (like `['videosById', { from: 0, to: 10 }, 'name' ]` where `videosById` is not in cache

fixes #779

### DIFF
--- a/test/data/LocalDataSource.js
+++ b/test/data/LocalDataSource.js
@@ -96,7 +96,7 @@ LocalSource.prototype = {
                 observer.onCompleted();
             }
 
-            if (wait === false) {
+            if (wait === undefined) {
                 exec();
             } else {
                 setTimeout(exec, wait);
--- a/test/data/LocalDataSource.js
+++ b/test/data/LocalDataSource.js
@@ -10,7 +10,7 @@ var LocalSource = module.exports = function(cache, options) {
         onGet: noOp,
         onSet: noOp,
         onResults: noOp,
-        wait: false,
+        // wait: undefined,
         materialize: false
     }, options);
     this._missCount = 0;
@@ -61,7 +61,7 @@ LocalSource.prototype = {
                 observer.onNext(output);
                 observer.onCompleted();
             }
-            if (wait === false) {
+            if (wait === undefined) {
                 exec();
             } else {
                 setTimeout(exec, wait);
--- a/test/data/LocalDataSource.js
+++ b/test/data/LocalDataSource.js
@@ -35,7 +35,7 @@ LocalSource.prototype = {
         var miss = options.miss;
         var onGet = options.onGet;
         var onResults = options.onResults;
-        var wait = +options.wait;
+        var wait = options.wait;
         var errorSelector = options.errorSelector;
         return Rx.Observable.create(function(observer) {
             function exec() {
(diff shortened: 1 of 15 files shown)


test/data/LocalDataSource.js
```diff
@@ -61,10 +61,10 @@ LocalSource.prototype = {
                 observer.onNext(output);
                 observer.onCompleted();
             }
-            if (wait > 0) {
-                setTimeout(exec, wait);
-            } else {
+            if (wait === false) {
```
jhusain: Agree with the decision to make "0" mean setTimeout(0). However would've used null/undefined to indicate no wait. That way the member could just be optional, and if not included, there would be no async.

asyncanup: thanks! points from discussion with @jhusain :

it did already was no wait if you missed the property (because wait gets defaulted to `false` earlier), but that did leave out the case where wait is explicitly sent in as undefined. agreed that missing the property should behave the same as setting it to undefined, so now:
using `undefined` to indicate no wait. not caring about `null` or other possible types at this point (this is test helper code).

PR #902, opened 2017-10-11
Adds reference emission flag

steveorsomethin:
Note for observers: this is a private, _temporary_ change to accommodate an internal need at Netflix, and this behavior will not be ported to 2.x+.


jhusain:
LGTM.

PR #903, opened 2017-10-19
protect against model.invalidate(undefined) and model.invalidate([])

asyncanup:
from blowing away the entire cache

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -226,7 +226,7 @@ Model.prototype.invalidate = function invalidate() {
     args = [];
     while (++argsIdx < argsLen) {
         args[argsIdx] = pathSyntax.fromPath(arguments[argsIdx]);
-        if (!Array.isArray(args[argsIdx])) {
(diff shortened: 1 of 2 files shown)


benmj:
👏

jhusain:
Worth discussion: Is TypeError the correct error to throw here? We'd like to move to nominal path types - ideally created via constructor - so that we don't have to keep running validation logic at every tier. We've always said that empty paths are invalid, because the root must be a Object. Under the circumstances should we consider an empty path a TypeError. Not sure if this is a misuse of TypeError. Certainly wouldn't be if we had a Path constructor so maybe we should wait until then to use TypeError?

asyncanup:
@jhusain fine with making this a TypeError. Also cool with making this a `InvalidPathError` for that matter :)
Will let you make the final call.

asyncanup:
@jhusain going to merge this, keeping existing `throw new Error` for now, instead of `TypeError` or `InvalidPathError`. Will make a new PR with your decision on that one once you're back

PR #905, opened 2017-10-24
expires: 0 works now

asyncanup:


jhusain:
LGTM.

lib/support/incrementVersion.js
```diff
@@ -1,4 +1,8 @@
 var version = 1;
 module.exports = function incrementVersion() {
+    // console.log(new Error().stack);
```
steveorsomethin: Minor: lingering comment

asyncanup: thanks for spotting!

steveorsomethin: Otherwise LGTM. Feel free to self-merge.

PR #907, opened 2017-12-02
Fixes #838 (Incorrect stack traces) and abstracts error type creation

### LINKED ISSUE #838: Errors have incorrect stack traces
I think the [claim made in comments](https://github.com/Netflix/falcor/blob/master/lib/errors/MaxRetryExceededError.js#L13-L14) is not actually true. I consistently see stack traces representing the source file of the error class, and the chain of modules that `require`d the error class itself, rather than runtime information.

In a REPL:

```
> const MaxRetryExceededError = require('falcor/lib/errors/MaxRetryExceededError')
undefined
> new MaxRetryExceededError().stack
'MaxRetryExceededError\n    at Object.<anonymous> (/Users/jon/dev/project/node_modules/falcor/lib/errors/MaxRetryExceededError.js:15:35)
  at Module._compile (module.js:541:32)
  at Object.Module._extensions..js (module.js:550:10)
  at Module.load (module.js:456:32)
  at tryModuleLoad (module.js:415:12)
  at Function.Module._load (module.js:407:3)
  at Module.require (module.js:466:17)
  at require (internal/module.js:20:19)
  at repl:1:13
  at REPLServer.defaultEval (repl.js:272:27)'
```

If I tweak the implementation somewhat I can achieve a correct stack trace:

```js
function MaxRetryExceededError() {
    Error.call(this);
    Error.captureStackTrace(this, this.constructor);
}

require('util').inherits(MaxRetryExceededError, Error);
```

```
MaxRetryExceededError: The allowed number of retries have been exceeded.
  at repl:1:1
  at REPLServer.defaultEval (repl.js:272:27)
```

andyfleming:
This change should provide a proper stack trace when errors are thrown (#838).

There was also some inconsistent, duplicated code between the error types. I abstracted it into a single function to create error types so it can be documented, tested, and maintained consistently. (It is not currently directly tested, but has full coverage through use cases).

### DIFF
--- a/lib/errors/createErrorType.js
+++ b/lib/errors/createErrorType.js
@@ -1,0 +1,67 @@
+var util = require("util");
+
+/**
+ * Creates a new type of error.
+ *
+ * @param {string} name - The error type name
+ * @param {string} message - The error message
+ * @param {string[]|function} additionalKeys - Key names to set from arguments (or function to manually do so)
+ *
+ * @returns {function} Returns an Error class
+ *
+ * @private
+ */
+function createErrorType(name, message, additionalKeys) {
+
+    // Create the error type
+    var NewErrorType = function() {
+
+        // Call the "super" constructor
+        Error.call(this);
+
+        // Capture a proper stack trace
+        Error.captureStackTrace(this, this.constructor);
+
+        // Set the message of our error type
+        this.message = message;
+
+        // If the additionalKeys argument provided was a function, call it with the instance and arguments
+        // (this allows additional logic like applying defaults)
+        if (typeof additionalKeys === "function") {
+            additionalKeys(this, arguments);
+        }
+
+        // Otherwise, if arguments were passed and this error type allows additional keys to be set, set them
+        if (arguments.length > 0 && Array.isArray(additionalKeys)) {
+
+            // For each additional key allowed...
+            additionalKeys.forEach(function(key, i) {
+
+                // Check if that argument was passed, and if so, set that property of this instance
+                if (typeof arguments[i] !== "undefined") {
+                    this[key] = arguments[i];
+                }
+            });
+        }
+    };
+
+    // Inherit from the standard error type
+    util.inherits(NewErrorType, Error);
+
+    // Override the name of the function
+    var descriptor = Object.getOwnPropertyDescriptor(NewErrorType, "name");
+    descriptor.value = name;
+    Object.defineProperty(NewErrorType, "name", descriptor);
+
+    // Override the name on the prototype
+    NewErrorType.prototype.name = name;
+
+    // provide a .is() method for checking the error type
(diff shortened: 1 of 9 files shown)


lib/errors/createErrorType.js
```diff
@@ -0,0 +1,67 @@
...
+function createErrorType(name, message, additionalKeys) {
+
+    // Create the error type
+    var NewErrorType = function() {
+
+        // Call the "super" constructor
+        Error.call(this);
+
+        // Capture a proper stack trace
+        Error.captureStackTrace(this, this.constructor);
```
steveorsomethin: This is unfortunately a v8-only API. You could feature detect however, and perhaps fall back to
```(new Error()).stack``` as before.

File: lib/errors/createErrorType.js [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+ * @param {string[]|function} additionalKeys - Key names to set from arguments (or function to manually do so)
+ *
+ * @returns {function} Returns an Error class
+ *
+ * @private
+ */
+function createErrorType(name, message, additionalKeys) {
+
+    // Create the error type
+    var NewErrorType = function() {
```
steveorsomethin: I can see where you're going with the Error class centralization, though there's a (small) drawback to this approach. Even with setting the name as you are, you'll end up with NewErrorType as the error name in the console. The stack will have the right name, but I'm afraid it will negatively impact at-a-glance debugging.

Very sorry to do this to you, but would you mind re-decentralizing the classes?

lib/errors/createErrorType.js
andyfleming: Oh, didn't even realize! If there's not any advantage to `.captureStackTrace` I would probably opt to keep it simple with `(new Error()).stack`.

lib/errors/createErrorType.js
andyfleming: No problem. I thought I had made it safely through renaming the error type but you're right that it shows incorrectly in the console. Bummer. Are you confident there's no way around this or is it worth a little more research on my end?

lib/errors/createErrorType.js
steveorsomethin: It could be useful, just when feature detected. Although, without a repro for the originating issue it's hard to know. Including the constructor stack frame, while annoying isn't a big deal. Including an incorrect stack as in the issue definitely is, but I wonder if it was a REPL only issue.

lib/errors/createErrorType.js
steveorsomethin: Nothing is turning up when I go hunting for ways to change that, unfortunately.

andyfleming:
@steveorsomethin Thanks for the feedback. I've opened a fresh PR, #908, approaching this how we discussed above.

(Closing this one.)

PR #908, opened 2017-12-20
Fixes #838 (Incorrect stack traces)

### LINKED ISSUE #838: Errors have incorrect stack traces

andyfleming:
This pull request fixes #838 (and replaces #907).

I've consistently applied the approach described here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#ES5_Custom_Error_Object

It provides a proper stack trace, the correct name, computes to true for `instanceof Error`, and should be widely compatible.

There is a single function `applyErrorPrototype` which abstracts some of the error prototype configuration without breaking any of the above features.


asyncanup:
Thanks for the PR, @andyfleming
Going to take this up as part of an effort to implement a consistent error handling strategy across the board.

andyfleming:
No problem. Sounds good @asyncanup.

lib/errors/applyErrorPrototype.js
```diff
@@ -0,0 +1,19 @@
...
+        writable: true,
+        configurable: true
+        }
+    });
+
+    if (Object.setPrototypeOf) {
+        Object.setPrototypeOf(errorType, Error);
+    } else {
+        // eslint-disable-next-line
+        InvalidModelError.__proto__ = Error;
```
asyncanup: this should be `errorType.__proto__`

lib/errors/InvalidSourceError.js
```diff
@@ -1,23 +1,34 @@
...
+    return instance;
 }
 
-// instanceof will be an error, but stack will be correct because its defined
-// in the constructor.
-InvalidSourceError.prototype = new Error();
-InvalidSourceError.prototype.name = NAME;
+applyErrorPrototype(InvalidSourceError);
+
 InvalidSourceError.is = function(e) {
```
asyncanup: this function can be deleted. `InvalidSourceError.is` is not used anywhere in source code.

lib/errors/MaxRetryExceededError.js
```diff
@@ -1,22 +1,34 @@
...
+    return instance;
 }
 
-// instanceof will be an error, but stack will be correct because its defined
-// in the constructor.
-MaxRetryExceededError.prototype = new Error();
-MaxRetryExceededError.prototype.name = NAME;
+applyErrorPrototype(MaxRetryExceededError);
+
 MaxRetryExceededError.is = function(e) {
```
asyncanup: delete the function and change `test/falcor/get/get.dataSource-only.spec.js` and `test/falcor/set/set.dataSource-only.spec.js` to use equality comparison with `MaxRetryExceededError.name` instead.

test/falcor/deref/deref.errors.spec.js
```diff
@@ -35,7 +35,7 @@ describe('Error cases', function() {
                     get([0, 0, 'item', 'title'])).
                     doAction(onNext, function(err) {
                         expect(onNext.callCount).to.equal(1);
-                        expect(err.message).to.equals(InvalidModelError.message);
+                        expect(err.message).to.equals("The boundPath of the model is not valid since a value or error was found before the path end.");
```
asyncanup: I think we should only check for `err.name` in tests, because that maps 1:1 with the error having the right `err.message`. simply delete the testing for `err.message` everywhere

test/get-core/deref.spec.js
```diff
@@ -80,8 +80,12 @@ describe('Deref', function() {
         })._derefSync(['videos', 0]);
 
         var res = model._getPathValuesAsJSONG(model, [['summary']], [{}]);
-        expect(res.criticalError.name).to.equals(BoundJSONGraphModelError.name);
-        expect(res.criticalError.message).to.equals(BoundJSONGraphModelError.message);
+        expect(res.criticalError.name).to.equals("BoundJSONGraphModelError");
+        expect(res.criticalError.message).to.equals(
```
asyncanup: same comment as above, delete checks for errors' `.message` property

asyncanup:
@andyfleming made some review comments^

test/falcor/deref/deref.errors.spec.js
andyfleming: So change

```js
expect(err.message).to.equals("The boundPath of the model is not valid since a value or error was found before the path end.");
```

to

```js
expect(err.message)
```

?


Or delete it completely?

asyncanup: delete completely

lib/errors/InvalidSourceError.js
```diff
@@ -1,23 +0,0 @@
-var NAME = "InvalidSourceError";
```
asyncanup: @andyfleming why did this file get deleted? makes the tests fail too!

andyfleming:
Closing. Changes have been pulled into #941.

PR #913, opened 2018-01-25
fixes #911 - "npm run dist" was breaking due to breaking api change in del@3.0.0

### LINKED ISSUE #911: Recent upgrades to dev dependencies is causing dist to silently fail
At least, I am speculating that the upgrades are the root cause. Running ```npm run dist``` now only cleans the dist directory, and the build task simply silently exits.

asyncanup:


jhusain:
LGTM

PR #914, opened 2018-01-26
add prepare script to ensure its run before every publish

asyncanup:


sdesai:
Is npm publish the right phase to do this? Is there value in having the ci build do this so the state of github dist matches the source (e.g for folks pulling straight from github)

Just making sure we’ve considered that aspect and rule it in/out explicitly.

lrowe:
Do we even need the ci committing dist files to git? Seems worth checking if this is still necessary as - npm/npm#3055 seems to suggest this is now fixed and a dependency on a git url would result in npm prepare being run and the dist generated.

lrowe:
If it is still a problem / too unwieldy perhaps we could configure ci publish prerelease versions to npm every merge to master (with <version>-<gitsha> version strings) instead of checking the build files back into git.

asyncanup:
@sdesai 
i thought https://github.com/Netflix/falcor/blob/master/.travis.yml already takes care of making a new commit with updated dist/ and doc/ folders when merging to master

asyncanup:
looks like it
screen shot 2018-01-26 at 12 19 53 pm

Issue #915, opened 2018-02-06
.batch resets maxRetries


quantumew:
I would like to build a batching Falcor model with a non-default maxRetries. I do the following

`const model = new falcor.Model({ maxRetries: 7, ... }).batch(..)`

Unfortunately this resets my maxRetries back to the default 3 due to the call to `this._clone`, which then instantiates the model passing in `this`, maxRetries is mapped to `this._maxRetries` and thus gets reverted to the default.

Looks like the fix would be mapping options._maxRetries as well in the constructor but perhaps this is intentional.

Is this expected behavior? Seems wrong to me but perhaps I am overlooking something.

asyncanup:
yeah that looks like a bug. will look into it and see what's going on

quantumew:
v1.1.0 fixes this. Thanks!

quantumew:
@asyncanup Just updated to v2.0.1 and it is great, except this issue has returned. Guessing the fix never got merged down. Didn't confirm that, just applied the same fix. See #932

quantumew:
@asyncanup Thanks for the release!!

asyncanup:
👍

PR #917, opened 2018-02-06
Fix max retries on clone

quantumew:
https://github.com/Netflix/falcor/issues/915

No clue if you guys take pull requests but this appears to fix the issue above.

For ease of reading at a glance I will reiterate what this is fixing. batch calls this._clone, _clone instantiates Model with `this`, the instance of the model does not have a property maxRetries but rather _maxRetries, therefore maxRetries gets reset to the default if you do the following

`new falcor.Model({maxRetries: 10, ... }).batch(...)`

Also I can write a test for this if you'd like

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -90,7 +90,7 @@ function Model(o) {
     if (typeof options.maxRetries === "number") {
         this._maxRetries = options.maxRetries;
     } else {
-        this._maxRetries = options.maxRetries || Model.prototype._maxRetries;
+        this._maxRetries = options._maxRetries || Model.prototype._maxRetries;
     }
 
     if (typeof options.collectRatio === "number") {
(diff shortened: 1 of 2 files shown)


asyncanup:
@quantumew 
good catch and great fix. could you also add a test case for this?

quantumew:
@asyncanup Thanks for the fast response. ebce4333fc29412a82b5ba99e1c873b34d74f8cb

quantumew:
@asyncanup Sorry to bother, do you have a timeline for this to be released? Just curious, thanks again for the time! Let me know if I am missing anything!

asyncanup:
LGTM

asyncanup:
will wait a day to let other reviewers review this (added them to the PR), and then merge into master. 
in terms of going out on npm as a new version, that can take a few days

quantumew:
@asyncanup Awesome, thank you sir!

jhusain:
LGTM

PR #918, opened 2018-02-10
new hook-in for falcor legacy polyfills

asyncanup:


lib/get/get.js
```diff
@@ -63,7 +63,13 @@ module.exports = function get(walk, isJSONG) {
                     criticalError: new BoundJSONGraphModelError()
                 };
             }
-            currentCachePosition = getCachePosition(model, boundPath);
+
+            // using _getOptimizedPath because that's a point of extension
+            // for polyfilling legacy falcor
+            optimizedPath = model._getOptimizedBoundPath();
```
michaelbpaulson: +1, good hook.

lib/Model.js
```diff
@@ -141,6 +141,16 @@ Model.prototype._collectRatio = 0.75;
  */
 Model.prototype.get = require("./response/get");
 
+Model.prototype._getOptimizedBoundPath = function _getOptimizedBoundPath() {
```
michaelbpaulson: Don't we have a copy already?

asyncanup: Copy of what?

michaelbpaulson: https://github.com/Netflix/falcor/blob/master/lib/support/array-clone.js

michaelbpaulson: I would imagine this function being a touch more like 
```
const clone = require('./path/to/array-clone');

Model.prototype._getOptimizedBoundPath = function _getOptimizedBoundPath() {
    return clone(this._path);
}
```

asyncanup: ah cool

michaelbpaulson:
lgtm

PR #920, opened 2018-02-19
Response stuffing

steveorsomethin:
This PR accomplishes two things:

It fixes a bug where whole branches were merged from jsonGraph responses, leading to nodes to be merged without also receiving their metadata such as $_absolutePath.

Adds a (for now) private model flag, _useServerPaths which can enable folks to opt into response stuffing. I opted for a private flag for now as I see this as a pretty major foot gun.

### DIFF
--- a/lib/request/GetRequestV2.js
+++ b/lib/request/GetRequestV2.js
@@ -105,7 +105,7 @@ GetRequestV2.prototype = {
                             currentCacheVersion.setVersion(currentVersion);
                             var mergeContext = {hasInvalidatedResult : false};
 
-                            var pathsErr = model._useServerPaths && data.paths === undefined ?
+                            var pathsErr = model._useServerPaths && data && data.paths === undefined ?
                                 new Error("Server responses must include a 'paths' field when Model._useServerPaths === true") : undefined;
 
                             if (!pathsErr) {
--- a/lib/request/GetRequestV2.js
+++ b/lib/request/GetRequestV2.js
@@ -104,14 +104,21 @@ GetRequestV2.prototype = {
                             var currentVersion = incrementVersion.getCurrentVersion();
                             currentCacheVersion.setVersion(currentVersion);
                             var mergeContext = {hasInvalidatedResult : false};
-                            self._merge(rPaths, err, data, mergeContext);
+
+                            var pathsErr = model._useServerPaths && data.paths === undefined ?
+                                new Error("Server responses must include a 'paths' field when Model._useServerPaths === true") : undefined;
+
+                            if (!pathsErr) {
+                                self._merge(rPaths, err, data, mergeContext);
+                            }
+
                             // Call the callbacks.  The first one inserts all
                             // the data so that the rest do not have consider
                             // if their data is present or not.
                             for (i = 0, len = callbacks.length; i < len; ++i) {
                                 fn = callbacks[i];
                                 if (fn) {
-                                    fn(err, data, mergeContext.hasInvalidatedResult);
+                                    fn(pathsErr || err, data, mergeContext.hasInvalidatedResult);
                                 }
                             }
                             currentCacheVersion.setVersion(null);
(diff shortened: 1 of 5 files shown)


test/set/edge-cases.spec.js
```diff
@@ -42,7 +42,7 @@ describe("Special Cases", function() {
         var edgeCaseCache = {
             jsonGraph: {
                 user: {
-                    name: "Jim",
+                    name: {$type: $atom, value: "Jim"},
```
steveorsomethin: This test was actually testing for buggy behavior before. It was asserting that the name field wasn't getting wrapped, which was wrong.

lib/support/mergeJSONGraphNode.js
```diff
@@ -192,7 +192,7 @@ module.exports = function mergeJSONGraphNode(
         }
     }
     else if (node == null) {
-        node = insertNode(message, parent, key, undefined, optimizedPath);
+        node = insertNode({}, parent, key, undefined, optimizedPath);
```
asyncanup: i don't totally understand this part. what was `message` about, and why is it now replaced with an empty object?

mattflix:
Is anybody looking at this PR?

The Windows apps team is waiting on it.

lib/request/GetRequestV2.js
```diff
@@ -104,14 +104,21 @@ GetRequestV2.prototype = {
                             var currentVersion = incrementVersion.getCurrentVersion();
                             currentCacheVersion.setVersion(currentVersion);
                             var mergeContext = {hasInvalidatedResult : false};
-                            self._merge(rPaths, err, data, mergeContext);
+
+                            var pathsErr = model._useServerPaths && data.paths === undefined ?
+                                new Error("Server responses must include a 'paths' field when Model._useServerPaths === true") : undefined;
```
lrowe: Is data guaranteed to be an object when err is an error?

lib/request/GetRequestV2.js
```diff
@@ -190,7 +197,7 @@ GetRequestV2.prototype = {
         model._path = emptyArray;
 
         // flatten all the requested paths, adds them to the
-        var nextPaths = flattenRequestedPaths(requested);
+        var nextPaths = model._useServerPaths ? data.paths : flattenRequestedPaths(requested);
```
lrowe: if requested paths need flattening might data.paths need flattening too?

lib/support/mergeJSONGraphNode.js
lrowe: It looks like message is the current JsonGraphNode of the response JsonGraph.

test/falcor/get/get.dataSource-only.spec.js
```diff
@@ -275,6 +275,89 @@ describe('DataSource Only', function() {
...
+    it('should honor response-stuffed paths with _useServerPaths == true.', function(done) {
+        var onGet = sinon.spy();
+        var source = new LocalDataSource(cacheGenerator(0, 2), {
+            onGet: onGet,
+            wait: 100,
+            onResults: function(data) {
+                data.paths = [
+                    ['videos', 0, 'title'],
+                    ['videos', 1, 'title']
+                ];
```
lrowe: I imagine this might be collapsed server side to `[['videos', { from: 0, to: 1 }, 'title']]`.

lib/support/mergeJSONGraphNode.js
jhusain: This change creates a new node instead of assuming that the node from the server is correctly formed. This prevents extra data stuffed into a node from the server from piggybacking into the cache.

lib/request/GetRequestV2.js
steveorsomethin: It doesn't, only because it's really just removing an outer wrapped on requested that shouldn't be present on the envelope paths.

test/falcor/get/get.dataSource-only.spec.js
steveorsomethin: Not a problem, as tons of tests already cover the case of complex keys (and they remain complex keys in _merge).

lib/request/GetRequestV2.js
steveorsomethin: Will add a null check just in case.

lib/support/mergeJSONGraphNode.js
asyncanup: perfect, this solves the issue: "model.set() claims object supplied in arguments" as well!

PR #922, opened 2018-02-22
fix branch name used for deploying docs

asyncanup:


steveorsomethin:
LGTM, CI is still running though.

PR #927, opened 2018-04-02
Update homepage-getting-started.md to match README

spadin:
Adds a line in `index.js` for serving static files from current directory so reader can visit `localhost:3000/index.html` in the next step.

Assuming this is the right file to update code sample at https://netflix.github.io/falcor/


asyncanup:
Thanks!

PR #930, opened 2018-04-23
Properly handle outerResults as an optional argument

emilgoldsmith:
Hey, sorry for the urgent flag, but this is an issue that is keeping my team from using Falcor v2 and it's a very simple fix so would be great to have it merged in and published quickly if possible.

We haven't completely isolated a fully reproducible case, but it has something to do with refs being followed, and we've been getting the following error message:
screenshot from 2018-04-23 10-34-38

The below small fix works though (tried changing it directly in my node_modules), and also seems it should be there as in `followReference` you explicitly pass `null` as the `outerResults` argument.

Hope this can be resolved quickly, thanks!

### DIFF
--- a/lib/get/onValue.js
+++ b/lib/get/onValue.js
@@ -132,6 +132,7 @@ module.exports = function onValue(model, node, seed, depth, outerResults,
           }
         }
     }
-
-    outerResults.hasValues = hasValues;
+    if (outerResults) {
+        outerResults.hasValues = hasValues;
+    }
 };


asyncanup:
Hard to accept this without a reproducible test case. Would it be possible to provide that? Or maybe even pointers as to how to hit this error.

asyncanup:
Essentially, we don't expect `outerResults` to be optional. If you're seeing that as null, then something else must be going on as the root cause.

emilgoldsmith:
Okay, I think I know how to reproduce it, I'll try and create a test repo in the near future (I'm busy in the next few days though). As of right now we also just made my fork our dependency until a fix is merged in here so we're not in such a rush.

asyncanup:
@emilgoldsmith closing this one out while you work on the PR with a reproducible test. thanks!

emilgoldsmith:
@asyncanup finally found some time to put in proper work on this. Here's a reproducible test case. Tried to make it as simple as possible to isolate the bug as much as possible:

https://github.com/emilgoldsmith/falcor-bug

Can we please reopen this? I'm not sure if we need to discuss a better solution than the one I hacked, but it'd be great to have a solution soon so we don't have to keep maintaining a fork of Falcor

asyncanup:
Reopened, looking into it

emilgoldsmith:
Hey, is there anything I can do to help out with this? @asyncanup. Would be great to get it fixed

jcranendonk:
@emilgoldsmith I'm looking into it, should have a PR with code test up soon.

emilgoldsmith:
Thanks @jcranendonk !

jcranendonk:
Closing, included in #947.

Issue #926, opened 2018-03-22
Assign to read only property


ghost:
Hello, here's something which confused me.

```
// falcor/lib/errors/InvalidDerefInputError.js

function InvalidDerefInputError() {
    this.message = MESSAGE;
    this.stack = (new Error()).stack;
}

// instanceof will be an error, but stack will be correct because its defined in the constructor.
InvalidDerefInputError.prototype = new Error();
InvalidDerefInputError.prototype.name = NAME;
InvalidDerefInputError.name = NAME;
InvalidDerefInputError.message = MESSAGE;
```

Now we can see there's a operation that assign to property name of InvalidDerefInputError, which is a read only property.
It works fine in normal mode but will throw err in strict mode.
Besides, is it a meaningful operation?Cause the function is already had a name...

Hope someone can help me out, thanks a lot :)

asyncanup:
I'm guessing that the original intent was to make the function keep its name even after being minified and mangled, as you can see it loses its full name in `dist/falcor.all.min.js`.
But I also realize that what we actually care about is just that the error instance retains the name, which is ensured by `InvalidDerefInputError.prototype.name = NAME` anyway, even when code is mangled. So that's good.

The only place where we do need `InvalidDerefInputError.name` to be the full function name is in the tests (`test/falcor/deref/deref.errors.spec.js`), but that test picks the error class from the actual file using node require, and not from minified build. Besides, falcor doesn't externally export the error class anyway. So we're good on that front too.

Long story short, I agree with the issue raised here and the corresponding fix to remove this name assignment line from all error types. 

@jhusain @lrowe @sdesai thoughts?

jcranendonk:
Addressed in https://github.com/Netflix/falcor/commit/896d2d7f543a4361622155b3f0fce50ef2578675

PR #931, opened 2018-04-26
Hotfix 1.x

ggd543:
#926


asyncanup:
Instead of commenting, you should delete the line outright, and delete it from the other Error classes as well.
That is, if the issue and PR does get accepted, we would want it to be comprehensive.

asyncanup:
Merged via #938 
Thanks @ggd543 !

PR #932, opened 2018-05-23
Issue #915 reintroduced in v2 of Falcor

quantumew:
Due to incorrect reference to internal property of _maxRetries when
cloning a Falcor model instance, which occurs internally when enabling
batching, maxRetries always gets reset when enabling batching.

Originally fixed in PR #917

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 
 node_js:
-    - "4"
     - "6"
+    - "8"
 
 branches:
     only:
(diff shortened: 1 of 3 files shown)


quantumew:
Note on the build failure, your travis build is installing the latest npm (now v6.0.1) and testing with node v4 which is EOL. It appears npm 6 simply does not run in node 4. So to get it to pass I would have to either fix the npm version to npm@5 or update versions of node tested in. I will for now update to test in v6 and v8. Please comment if this is not an acceptable change

asyncanup:
LGTM!

asyncanup:
Merging for now, open to reverting if other reviewers have an issue with it.

PR #933, opened 2018-06-05
Fix model get dispose

asyncanup:
Currently, dispose a model.get().subscribe() does not abort the inner XHR calls made by that model.get
This PR fixes that by threading the model get dispose through correctly to the get request's dispose.

This is the relevant commit that makes the fix: https://github.com/Netflix/falcor/commit/4c9ca5d8
Tests coming shortly.


jhusain:
This looks good pending test cases for the synchronous and asynchronous cases.

asyncanup:
@jhusain 
there was already a test case for synchronous case, i added new ones for the asynchronous case.

PR #937, opened 2018-07-26
fix(_setValueSync): fixed _setValueSync to use _getValueSync right

asyncanup:


lib/set/sync.js
```diff
@@ -35,5 +35,5 @@ module.exports = function setValueSync(pathArg, valueArg, errorSelectorArg, comp
 
     this._syncCheck("setValueSync");
     setPathValues(this, [value]);
-    return this._getValueSync(this, value.path).value;
+    return this._getValueSync(value.path);
```
lrowe: Looking at lib/get/sync.js perhaps this should instead be `return getValueSync(this, value.path).value;` to avoid the duplicate path type checking.

Also it looks like only _getValueSync are not _setValueSync are not symmetrical as only _getValueSync prefixes with the deref this._path.

asyncanup: agreed to the point about duplicated path type checking, making the change.

to the point of symmetry, i think `_geValueSync` and `_setValueSync` are symmetrical. 
this works:
```
const falcor = require('..');
const assert = require('assert');

const model = falcor({ _path: ['some'], cache: { some: { thing: 'something'}}});
assert.deepEqual(model._path, ['some']);
model._setValueSync([ 'thing' ], 'something else');
assert.equal(model._getValueSync('thing'), 'something else');
console.log('success!');
```

lrowe: Digging deeper it looks like the bound path is added by the underlying setPathValues (but not in getValueSync, which requires _getValueSync to add it). That means we'd need to replicate the bound path concat code from _getValueSync into _setValueSync too:
```
    var path = value.path;
    if (this._path.length) {
        path = this._path.concat(path);
    }
    return getValueSync(this, path).value;
```

asyncanup: actually about the first point, I'm now thinking of leaving `this._getValueSync` in there instead of `getValueSync`, because for `_setValueSync`, the bound path is calculated inside `setPathValues` whereas `getValueSync` expects the bound path.

asyncanup: so it would involve multiple path syntax checks anyway, unless i move the path check out of `setPathValues` and put it into `_setValueSync`, but that may have bigger implications. doesn't seem worth it at this point

PR #945, opened 2018-08-22
fix minor syntax error in README.md

ntower:
the example code had a syntax error due to an errant semicolon inside of an object literal


NickHeiner:
Thanks!

PR #947, opened 2018-10-04
Fixing double ref JSONG issue

jcranendonk:
In the onNext consistency fix, a null check was omitted which caused `onValue` to fail to handle outerValues being unset. This situation only occurs in JSONG mode when a ref to a ref is followed from `followReference`, which was not covered by a code test.

This PR adds a code test for that scenario, and pulls in the fix from #930.


emilgoldsmith:
Thank you!

PR #949, opened 2018-10-29
Deduplication of multiple overlapping requests

jcranendonk:
When RequestQueue deduplicates multiple get requests with intersecting path sets that include ranges or key sets, a situation can occur where remaining paths are generated with an `undefined, undefined` prefix due to mismatching lengths between the remaining/optimized path arrays and the depthDifferences array. This causes unexpected NullInPathErrors. Removing the depthDifferences array and calculating path length difference directly resolves this issue.

An additional issue was found in `complement()`, where it called `findPartialIntersections()` with the incorrect subtree when an optimized path was longer than its corresponding requested path. The issue was resolved by descending into the subtree to the correct node in that scenario. The two methods were slightly reworked and cleaned up, and unit tests were added.


michaelbpaulson:
It looks good (I think).  Its been a bit.  I wonder why I used a depth array... it does not seem to make much sense.  :)

sdesai:
We should probably have a unit test for complement() for this use case shouldn't we? The integration test is good.

jcranendonk:
@sdesai good point; now that `complement()` no longer relies on the parameter, it makes sense to test it directly.

NickHeiner:
> This causes unexpected NullInPathErrors.

Are we seeing this in production? Do we have queries to track the error rate now and after this fix is applied?

NickHeiner:
Is the "requested" v. "optimized" path distinction just following refs?

lib/request/complement.js
```diff
@@ -100,34 +64,52 @@ module.exports = function complement(requested, optimized, depthDifferences, tre
...
+    var depthDiff = requestedPath.length - optimizedPath.length;
+
+    // Pre-descend into the sub tree (when the optimized path is longer than the requested path)
+    for (var i = 0; currentTree && i < -depthDiff; i++) {
+        currentTree = currentTree[optimizedPath[i]];
+    }
+
+    // No matching path, cannot dedupe
+    if (!currentTree) {
+        return [[], [optimizedPath], [requestedPath]];
```
NickHeiner: I presume this is just being consistent with the rest of the codebase, but I generally find objects to be a much more readable way of returning multiple values than arrays.

lib/request/complement.js
```diff
@@ -100,34 +64,52 @@ module.exports = function complement(requested, optimized, depthDifferences, tre
...
+ *  - the intersection of requested paths with currentTree
+ *  - the complement of optimized paths with currenTree
+ *  - the complement of corresponding requested paths with currentTree
  */
-function findPartialIntersections(requestedPath, optimizedPath, currentTree, depth, rCurrentPath, oCurrentPath, depthDiff) {
+function findPartialIntersections(requestedPath, optimizedPath, currentTree) {
+    var depthDiff = requestedPath.length - optimizedPath.length;
+
+    // Pre-descend into the sub tree (when the optimized path is longer than the requested path)
+    for (var i = 0; currentTree && i < -depthDiff; i++) {
```
NickHeiner: Minor: I prefer to declare `var i` outside of the `for` loop, to remind people that the variable is in scope for the entire function and not just the `for` loop.

lib/request/complement.js
```diff
@@ -100,34 +64,52 @@ module.exports = function complement(requested, optimized, depthDifferences, tre
...
+        currentTree = currentTree[optimizedPath[i]];
+    }
+
+    // No matching path, cannot dedupe
```
NickHeiner: Nit: I prefer to have fully formed English sentences as code comments.

```js
// There is no matching path, so deduping is impossible.
```

I'm not sure how others feel.

lib/request/complement.js
```diff
@@ -100,34 +64,52 @@ module.exports = function complement(requested, optimized, depthDifferences, tre
...
+    if (!currentTree) {
+        return [[], [optimizedPath], [requestedPath]];
+    }
+
+    if (depthDiff === 0) {
+        return findPartialIntersectionsRec(requestedPath, optimizedPath, currentTree, 0, [], [], 0);
+    } else if (depthDiff > 0) {
+        return findPartialIntersectionsRec(requestedPath, optimizedPath, currentTree, 0, arraySlice(requestedPath, 0, depthDiff), [], depthDiff);
+    } else { // depthDiff < 0
```
NickHeiner: I for one would not have a comment like this, because 1) I think the code is already pretty clear and 2) it's at risk of getting out of sync with the code. But it's fine to leave it in.

lib/request/complement.js
```diff
@@ -100,34 +64,52 @@ module.exports = function complement(requested, optimized, depthDifferences, tre
...
+        return findPartialIntersectionsRec(requestedPath, optimizedPath, currentTree, -depthDiff, [], arraySlice(optimizedPath, 0, -depthDiff), depthDiff);
+    }
+}
+
+function findPartialIntersectionsRec(
```
NickHeiner: It looks like `requestedPath` and `optimizedPath` are passed through the recursive calls unchanged. What would you think about defining `findPartialIntersectionRec` within `findPartialIntersection` and just accessing them through the closure?

lib/request/complement.js
```diff
@@ -211,3 +192,5 @@ function findPartialIntersections(requestedPath, optimizedPath, currentTree, dep
     return [intersections, oComplementPaths, rComplementPaths];
 }
 
+// Exported for unit testing.
+module.exports.findPartialIntersections = findPartialIntersections;
```
NickHeiner: You could lose the comment above and prevent accidental usage of this function by doing:

```js
module.exports.__test = {findPartialIntersections: findPartialIntersections};
```

lib/response/get/checkCacheAndReport.js
```diff
@@ -71,11 +70,7 @@ module.exports = function checkCacheAndReport(model, requestedPaths, observer,
     // 2.  The request if finished and the json key off
     // the valueNode has a value.
     if (progressive || ((progressive && results.hasValues || !progressive) && completed && valueNode !== undefined)) {
-        try {
-            observer.onNext(valueNode);
-        } catch (e) {
-            throw e;
-        }
+        observer.onNext(valueNode);
```
NickHeiner: The `try catch` was vacuous, right?

lib/request/complement.js
jcranendonk: Good point, I'll update.

test/integration/dedupe.spec.js
```diff
@@ -198,5 +200,22 @@ describe('Request deduping', function() {
             ]);
         }, done, partDone);
     });
+
+    it("deduplicates gets with overlapping ranges", function (done) {
+        var onGet = sinon.spy();
+        var model = new falcor.Model({ source: new LocalDataSource(cacheGenerator(0, 3), { wait: 0, onGet: onGet }) });
+
+        var partDone = after(3, function() {
```
NickHeiner: Since we already have `lodash` in this package, I'd just as soon use [`_.after`](https://lodash.com/docs/4.17.11#after) so we can cut down on having this extra dep. But I also see the appeal of consistency with the rest of the codebase. It's not a big deal either way.

lib/response/get/checkCacheAndReport.js
jcranendonk: It was. Originally, there was a `finally` as well, but it had been removed while leaving the try/catch in place.

test/integration/dedupe.spec.js
```diff
@@ -198,5 +200,22 @@ describe('Request deduping', function() {
...
+            expect(onGet.getCall(0).args[1]).to.deep.equal([["videos", 0, "title"]]);
```
NickHeiner: Do we really care what order the `onGet` calls were in? Wouldn't we be happy with these calls occurring in any order? If so, can we make a looser assertion that's less brittle?

lib/request/complement.js
jcranendonk: I considered it but it felt like too large a change initially. As it turns out, `complement()` is only used in one place, and the logic there could use a bit of a cleanup too. I'll update it.

test/internal/request/complement.spec.js
```diff
@@ -0,0 +1,113 @@
+var expect = require("chai").expect;
+var complement = require("../../../lib/request/complement");
+var findPartialIntersections = require("../../../lib/request/complement").findPartialIntersections;
+
+describe("complement", function() {
+    it("returns null if no paths are provided", function() {
+        expect(complement([], [], {})).to.be.null;
```
NickHeiner: Ending an assertion chain with a property accessor is dangerous because the following test would also pass:

```js
        expect(complement([], [], {})).to.be.nul;
```

I've been bitten by this in the past. To solve it, I recommend every assertion chain ending with a method call:

```suggestion
        expect(complement([], [], {})).to.equal(null);
```

test/internal/request/complement.spec.js
```diff
@@ -0,0 +1,113 @@
+var expect = require("chai").expect;
```
NickHeiner: As long as we're making a brand new file, why not do it in ES6?

test/internal/request/complement.spec.js
```diff
@@ -0,0 +1,113 @@
...
+    it("returns null if no paths are provided", function() {
+        expect(complement([], [], {})).to.be.null;
+    });
+
+    it("returns null if no deduping was possible", function() {
+        var paths = [["videos", 0, "title"]];
+        expect(complement(paths, paths, {})).to.be.null;
+    });
+
+    it("returns the complement and intersection consisting of paths that can be partially deduped", function() {
```
NickHeiner: Reminder to myself to ask in person what's going on in this test. 😄

test/internal/request/complement.spec.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+});
+
+describe("findPartialIntersections", function() {
+    var matchingPath;
+    var matchingPathTree;
+
+    beforeEach(function() {
+        matchingPath = ["lolomo", 123, "summary"];
+        matchingPathTree = { lolomo: { "123": { summary: null } } };
```
NickHeiner: The quotes are unnecessary around `"123"`. All object keys are strings.

```suggestion
        matchingPathTree = { lolomo: { 123: { summary: null } } };
```

Eslint has a rule to catch this. It would be great if it were enabled here.

test/integration/dedupe.spec.js
jcranendonk: Good catch, the `after` package is actually only used here. I'll remove it in lieu of lodash.

NickHeiner:
LGTM. But I don't have enough context to review deeply. @asyncanup should review before we merge.

lib/request/complement.js
```diff
@@ -54,30 +38,9 @@ module.exports = function complement(requested, optimized, depthDifferences, tre
...
         var subTree = tree[oPath.length];
 
-        // no deduping possible
-        if (!subTree) {
-            optimizedComplement[++complementLength] = oPath;
-            requestedComplement[complementLength] = rPath;
-            continue;
-        }
-        // fully deduped
-        if (hasIntersection(subTree, oPath, 0)) {
```
asyncanup: @jcranendonk is there any performance difference between running `hasIntersection` and `findPartialIntersections`? Might very well depend on their relative usage in apps..

lib/request/complement.js
```diff
@@ -100,34 +64,52 @@ module.exports = function complement(requested, optimized, depthDifferences, tre
...
+}
+
+function findPartialIntersectionsRec(
+    requestedPath,
+    optimizedPath,
+    currentTree,
+    depth,
+    rCurrentPath,
+    oCurrentPath,
+    depthDiff
```
asyncanup: `findPartialIntersectionsRec` has an inner `depthDiff` as well as an outer one. Can feel confusing to someone reading the code. If you're not modifying its value, might be worth just keeping `depthDiff` in the outer scope alone.

jcranendonk:
@nth "unexpected NullInPathErrors": Website UI has a work-around. No other platforms currently use this version. Several integration tests in TV UI failed due to this bug.

Optimized vs requested: if a path contains a `ref` which is in the local cache, it is dereferenced before the request is made. The dereferenced path is the 'optimized' path.

test/internal/request/complement.spec.js
```diff
@@ -0,0 +1,113 @@
...
+
+    it("returns null if no deduping was possible", function() {
+        var paths = [["videos", 0, "title"]];
+        expect(complement(paths, paths, {})).to.be.null;
+    });
+
+    it("returns the complement and intersection consisting of paths that can be partially deduped", function() {
+        var partialMatchingPath = ["videos", [0, 1], "title"];
+        var paths = [partialMatchingPath];
+        var pathTree = { "3": { videos: { "0": { title: null } } } };
```
asyncanup: can remove the quotes from keys here too. looks more readable to me that way

test/integration/dedupe.spec.js
jcranendonk: I'll see if I can do that without getting too complicated.

For reference, some functionality in Falcor is explicitly synchronous, such as returning cached data immediately. I don't believe that includes outgoing requests, however. (As far as Falcor knows, it's an implementation detail of DataSource implementations.)

test/internal/request/complement.spec.js
jcranendonk: Wasn't aware of that, thanks.

test/internal/request/complement.spec.js
```diff
@@ -0,0 +1,113 @@
...
+
+    it("returns the intersection consisting of paths that can be fully deduped", function() {
+        expect(findPartialIntersections(matchingPath, matchingPath, matchingPathTree)).to.deep.equal([
+            [matchingPath],
+            [],
+            []
+        ]);
+    });
+
+    describe("with optimized paths shorted than requested paths", function() {
```
asyncanup: `shorter than`

test/internal/request/complement.spec.js
jcranendonk: For posterity: `complement()` calculates which of the paths are contained in the passed-in tree, which is a representation of paths already being requested. 'Intersection' and 'complement' refer to paths that do and do not, respectively, appear in the tree.

test/internal/request/complement.spec.js
jcranendonk: Eslint is pretty out of date too. I'll see if I can update it.

lib/request/complement.js
jcranendonk: Source here: https://github.com/Netflix/falcor-path-utils/blob/master/lib/hasIntersection.js

The complexity of `hasIntersection` and `findPartialIntersections` looks essentially the same to me. The only difference is that the latter builds up the intersection/complement arrays, and updates the current path on recursion. In the case where `hasIntersection` would return true, I think exactly the same work is done.

lib/request/complement.js
asyncanup: my bad, should've been reading the code as a file. missed indentation in the diff view, and thought `findPartialIntersectionsRec` is defined _inside_ `findPartialIntersection`

test/integration/dedupe.spec.js
asyncanup: well since you're not enabling batching in the `model` here, it is expected that the requests will be made in sequence. i guess you could say that we don't explicitly say this in the contract between falcor and datasource, so it's just an _expectation_

lib/request/complement.js
jcranendonk: I did it 🎉

test/.eslintrc.yaml
```diff
@@ -0,0 +1,17 @@
+env:
+  es6: true
```
jcranendonk: Bring the goodness.

test/.eslintrc.yaml
```diff
@@ -0,0 +1,17 @@
...
+  mocha: true
+  node: true
+
+rules:
```
jcranendonk: Rules shamelessly copied from https://github.com/NetflixUI/nf-cl-logger

test/integration/dedupe.spec.js
jcranendonk: Decided not to mess with it right now. Call order is due to LocalDataSource implementation, which was specifically written for the test suite.

lib/request/complement.js
sdesai: It may have been an array for perf reasons (TV). @michaelbpaulson may know.

NickHeiner:
🍾 for ES6 in the tests.

PR #955, opened 2019-03-01
Bugfix: empty keyset handling

jcranendonk:
Empty keysets would result in paths with spurious keys of `"undefined"` to be generated for the requested/optimized missing path arrays during `walkPath`.


lib/get/onValue.js
```diff
@@ -20,7 +20,7 @@ module.exports = function onValue(model, node, seed, depth, outerResults,
     var i, len, k, key, curr, prev = null, prevK;
     var materialized = false, valueNode, nodeType = node && node.$type, nodeValue = node && node.value;
 
-    if (!node || node.value === undefined) {
+    if (nodeValue === undefined) {
```
jcranendonk: Little bit of cleanup.

lib/get/onValue.js
```diff
@@ -58,7 +58,10 @@ module.exports = function onValue(model, node, seed, depth, outerResults,
         }
     }
 
-    else {
+    else if (node && nodeType === undefined && nodeValue === undefined) {
```
jcranendonk: Required to ensure paths that terminate before a leaf (e.g. due to empty keysets) still include the key prior to the empty step.

lib/get/walkPath.js
```diff
@@ -16,58 +16,40 @@ module.exports = function walkPath(model, root, curr, path, depth, seed,
...
+    var isKeySet = keySet !== null && typeof keySet === "object";
+    var iteratorNote = isKeySet ? {} : undefined;
+    var key = isKeySet ? iterateKeySet(keySet, iteratorNote) : keySet;
+
+    // The walk is finished when:
+    // - there is no value in the current cache position
+    // - there is a JSONG leaf node in the current cache position
+    // - we've reached the end of the path
+    // - the current key set contains no keys
+    if (!curr || curr.$type || depth === path.length || (isKeySet && iteratorNote.done)) {
```
jcranendonk: Empty keysets terminate the walk early.

lib/get/walkPath.js
```diff
@@ -16,58 +16,40 @@ module.exports = function walkPath(model, root, curr, path, depth, seed,
...
-        if (key === null) {
-            // If the key is null and we are not at the end of a path, then
-            // throw an error.
-            if (depth < path.length) {
-                throw new NullInPathError();
-            }
+    var i;
 
-            // Else, we are at the end of a path, then just say next is current.
-            else {
```
jcranendonk: I don't believe it was possible to actually hit this branch.

lib/response/get/checkCacheAndReport.js
```diff
@@ -63,13 +62,10 @@ module.exports = function checkCacheAndReport(model, requestedPaths, observer,
...
-    // Which are under two conditions:
-    // 1.  This request is progressive
-    //
-    // 2.  The request if finished and the json key off
-    // the valueNode has a value.
-    if (progressive || ((progressive && results.hasValues || !progressive) && completed && valueNode !== undefined)) {
+    // Report locally available values if:
+    // - the request is in progressive mode, or
+    // - the request is complete and values were found
+    if (progressive || (completed && valueNode !== undefined)) {
```
jcranendonk: Unrelated to the PR but I've been eyeing this condition for cleanup. They're equivalent.

package.json
```diff
@@ -84,7 +84,7 @@
     "falcor-asap": "2.0.6",
     "falcor-json-graph": "1.1.7",
     "falcor-path-syntax": "0.2.4",
-    "falcor-path-utils": "0.5.1",
+    "falcor-path-utils": "0.7.2",
```
jcranendonk: Updated to pull in a [bugfix in iterateKeySet](https://github.com/Netflix/falcor-path-utils/commit/5e9d7e01849e9ab6b757627b8c701627e9d6b825).

test/get-core/edges.spec.js
```diff
@@ -19,7 +19,9 @@ describe('Edges', function() {
                     videos: {}
                 }
             },
-            cache: cacheGenerator(0, 1)
+            cache: cacheGenerator(0, 1),
+            requestedMissingPaths: [],
+            optimizedMissingPaths: []
```
jcranendonk: This tests for those spurious `"undefined"` missing paths. Prior to this bugfix, these arrays would contain `[["videos", "undefined", "title"]]`.

test/set/jsonGraphs/primitive.spec.js
```diff
@@ -49,10 +49,10 @@ describe("a primitive value", function() {
                 ])]
             );
         } catch(e) {
-            expect(e.message).to.equal("`null` is not allowed in branch key positions.");
+            errored = true;
+            expect(e).instanceOf(NullInPathError);
         }
 
-        errored = true;
         expect(errored).to.be.true;
```
jcranendonk: True equals true, marvelous. Moved the flag to the correct position.

lib/get/walkPath.js
```diff
@@ -16,58 +16,40 @@ module.exports = function walkPath(model, root, curr, path, depth, seed,
...
 
-        // The standard case, do the depth search into the cache.
-        else {
-            next = curr[key];
-            optimizedPath[optimizedLength] = key;
-            requestedPath[depth] = key;
-        }
+        fromReference = false;
+        optimizedPath[optimizedLength] = key;
+        requestedPath[depth] = key;
```
jcranendonk: This is where `key` would be `undefined` which eventually ends up in missing paths.

lib/get/walkPath.js
```diff
@@ -16,58 +16,40 @@ module.exports = function walkPath(model, root, curr, path, depth, seed,
...
-            }
+    var i;
 
-            // Else, we are at the end of a path, then just say next is current.
-            else {
-                next = curr;
-            }
+    // loop over every key in the key set
+    do {
+        if (key == null) {
```
jcranendonk: Using `==` because `key` can also be `undefined`, which I don't believe should be allowed either.

test/falcor/get/get.dataSource-and-cache.spec.js
```diff
@@ -133,51 +133,37 @@ describe('DataSource and Partial Cache', function() {
...
-                },
-                source: {
-                    get: onGet
-                }
+                cache: { b: {} },
+                source: new LocalDataSource({}, { onGet: onGet })
             });
 
-            var modelGet = model.get(['b', {to:0, from:1}]);
+            var modelGet = model.get(['b', { from: 1, to: 0 }, 'leaf']);
```
jcranendonk: Updated these two tests that—with a little bit of tweaking—catch the original bug. The original intent of the tests remains intact. The tests now also give proper feedback on failure.

PR #960, opened 2019-05-06
fix: typescript transpilation / assertion error in 2.9.x

rob-pocklington-a139759:
**Case in Point**
Using `falcor@^2.0.6` with Angular 6 with Typescript 2.9.2 when generating source maps fails with this message from Typescript transpilation: `Error: Debug Failure. False expression.`

**Issues fixed**
Because of the `a = b = c` structure used below (in `GetResponse.js` and `SetResponse.js`):
`var GetResponse = module.exports = function GetResponse(model, paths ...`
This triggers the following github issue: https://github.com/Microsoft/TypeScript/issues/24963.
   
As we are using Angular 6, it is undesirable to change Typescript versions, so i'm fixing the problem at the source.  There are other areas where this pattern exists in the **falcor** codebase, but this is the minimal change to solve our immediate issue.  Happy to update the other files if it's a requirement.

I have tested our Angular project with this branch and it works successfully.

### DIFF
--- a/lib/response/get/GetResponse.js
+++ b/lib/response/get/GetResponse.js
@@ -14,10 +14,8 @@ var getSize = require("./../../support/getSize");
  * @augments ModelResponse
  * @private
  */
-var GetResponse = module.exports = function GetResponse(model, paths,
-                                                        isJSONGraph,
-                                                        isProgressive,
-                                                        forceCollect) {
+var GetResponse = function GetResponse(model, paths, isJSONGraph,
+                                       isProgressive, forceCollect) {
     this.model = model;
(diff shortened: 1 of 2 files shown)


jcranendonk:
Looks good! If you'd like to make more PRs of this type, please feel free to. I can cut a release either after this PR, or wait until you make additional changes.

luke-rielley-a134226:
Hi @jcranendonk , thanks very much for approving this PR. Would you kindly be able to cut a release and push this to NPM? That would unblock us and enable us to get our Falcor implementation into production! Luke.

jcranendonk:
Released in 2.0.7.

PR #962, opened 2019-07-01
Add flags for expensive path algorithms

jcranendonk:
Added flags to disable request deduplication and path collapsing. These are expensive algorithms and we'd like to quantify their impact on users.

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -38,20 +38,24 @@ Model.ref = jsong.ref;
 Model.atom = jsong.atom;
 Model.error = jsong.error;
 Model.pathValue = jsong.pathValue;
+
 /**
  * This callback is invoked when the Model's cache is changed.
  * @callback Model~onChange
  */
 
- /**
- * This function is invoked on every JSONGraph Error retrieved from the DataSource. This function allows Error objects to be transformed before being stored in the Model's cache.
+/**
+ * This function is invoked on every JSONGraph Error retrieved from the DataSource. This function allows Error objects
+ * to be transformed before being stored in the Model's cache.
  * @callback Model~errorSelector
  * @param {Object} jsonGraphError - the JSONGraph Error object to transform before it is stored in the Model's cache.
  * @returns {Object} the JSONGraph Error object to store in the Model cache.
  */
 
- /**
- * This function is invoked every time a value in the Model cache is about to be replaced with a new value. If the function returns true, the existing value is replaced with a new value and the version flag on all of the value's ancestors in the tree are incremented.
+/**
+ * This function is invoked every time a value in the Model cache is about to be replaced with a new value. If the
+ * function returns true, the existing value is replaced with a new value and the version flag on all of the value's
+ * ancestors in the tree are incremented.
  * @callback Model~comparator
  * @param {Object} existingValue - the current value in the Model cache.
  * @param {Object} newValue - the value about to be set into the Model cache.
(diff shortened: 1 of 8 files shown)


lib/request/RequestQueueV2.js
```diff
@@ -63,36 +66,39 @@ RequestQueueV2.prototype = {
...
+                    // into the outgoing results.  Then our paths will be reduced
+                    // to the complement.
+                    if (results[0]) {
+                        rRemainingPaths = results[1];
+                        oRemainingPaths = results[2];
+                        disposables[disposables.length] = results[3];
+                        ++count;
+
+                        // If there are no more remaining paths then exit the loop.
+                        if (!oRemainingPaths.length) {
```
asyncanup: could it be useful to change request.add to potentially return `null` instead of empty array for `rRemainingPaths` and `oRemainingPaths`, just like you've done for them below.

lib/request/RequestQueueV2.js
```diff
@@ -63,36 +66,39 @@ RequestQueueV2.prototype = {
...
                 }
             }
 
-            // If there is a non sent request, then we can batch and leave.
+            // If there is an unsent request, then we can batch and leave.
             else {
                 request.batch(rRemainingPaths, oRemainingPaths, refCountCallback);
-                oRemainingPaths = [];
-                rRemainingPaths = [];
+                oRemainingPaths = null;
```
asyncanup: i understand that the reasoning behind doing this is to reduce garbage creation?

lib/request/flushGetRequest.js
```diff
@@ -21,42 +21,56 @@ module.exports = function flushGetRequest(request, listOfPaths, callback) {
...
 
-    // Take the pathMapTree and create the collapsed paths and send those
-    // off to the server.
-    var collapsedPaths = request._collasped = toPaths(pathMap);
-    var jsonGraphData;
+    if (model._enablePathCollapse) {
+        // Take the pathMapTree and create the collapsed paths and send those
+        // off to the server.
+        requestPaths = toPaths(request._pathMap);
+    } else if (listOfPaths.length === 1) {
```
asyncanup: what's the purpose behind differentiating between `listOfPaths` being single-element or multi-element array?

lib/request/flushGetRequest.js
```diff
@@ -21,42 +21,56 @@ module.exports = function flushGetRequest(request, listOfPaths, callback) {
...
+        requestPaths = listOfPaths[0];
+    } else {
+        requestPaths = Array.prototype.concat.apply([], listOfPaths);
```
asyncanup: what's wrong with sending `listOfPaths` itself to `model._source.get()`, instead of sending a copy?

lib/request/RequestQueueV2.js
jcranendonk: I'm unsure what you mean; `request.add` is not called when request deduplication is disabled. I made the switch to `null` assignments on lines 92-93 to avoid the allocation of unused empty arrays.

lib/request/RequestQueueV2.js
jcranendonk: Correct. The arrays are not used afterwards except to check they're empty, so these allocations were unnecessary.

lib/request/flushGetRequest.js
jcranendonk: This line performs a shallow flatten of `listOfPaths`:

`[[["todos", [0, 1], "description"]], [["items", [2, 3], "title"]]]`
=>
`[["todos", [0, 1], "description"], ["items", [2, 3], "title"]]`

Despite the name, `listOfPaths` is an Array of Arrays of path sets, while `DataSource#get` expected an Array of path sets.

lib/request/flushGetRequest.js
jcranendonk: Just a small optimization. When batch times are short or batching is disabled, or when request rates are low (relative to the batch timer), `listOfPaths` will contain only the single Array of path sets. Essentially, this is the happy path for the simple case.

lib/request/RequestQueueV2.js
asyncanup: aligned offline that gains from optimizing array allocations inside of `request.add` will be a larger undertaking.

lib/request/flushGetRequest.js
asyncanup: aligned offline. my point of confusion was that `requestPaths` seemed to be turned into different types in these conditional blocks. Jeroen explained how it's not. Would recommend adding a comment explaining the same thing to an inattentive observer.

asyncanup:
just adding tests would be great, otherwise looks good!

lib/Model.js
```diff
@@ -38,47 +38,63 @@ Model.ref = jsong.ref;
...
- * @param {?Model~onChange} options.onChange - a function called whenever the Model's cache is changed
- * @param {?Model~comparator} options.comparator - a function called whenever a value in the Model's cache is about to be replaced with a new value.
+ * @param {Options} [o] - a set of options to customize behavior
  */
 function Model(o) {
-
     var options = o || {};
     this._root = options._root || new ModelRoot(options);
     this._path = options.path || options._path || [];
-    this._scheduler = options.scheduler || options._scheduler || new ImmediateScheduler();
```
asyncanup: love removing this property given that it is indeed never used outside of initializing `_request` below, especially because there are many models derived from the root one, but almost never a new RequestQueue

test/Model.spec.js
```diff
@@ -347,101 +358,137 @@ describe("Model", function() {
...
+        const model = new MyModel();
         expect(model).toBeInstanceOf(MyModel);
-        var clonedModel = model.batch(100);
+        const clonedModel = model.batch(100);
         expect(clonedModel).toBeInstanceOf(MyModel);
     });
 
-    describe('JSON-Graph Specification', function() {
-        require('./get-core');
+    describe("algorithm options", () => {
```
asyncanup: instead of testing all algorithm options together, i'd prefer them tested separately, just because we may have more algorithm-related options later which may not be orthogonal to each other in the way these two are.

i propose the following structure for the test hierarchy here:
```
describe('Model...
    describe('disable path collapse algorithm'...
    describe('disable request de-duplication'...
```

test/internal/request/RequestQueue.get.spec.js
```diff
@@ -238,75 +244,175 @@ describe('#get', function() {
...
+        const disposable2 = queue.get([videos0, videos1], [videos0, videos1], zip);
         expect(queue._requests.length).toBe(2);
         expect(queue._requests[1].sent).toBe(true);
         expect(queue._requests[1].scheduled).toBe(false);
 
         disposable();
         disposable2();
     });
+
+    it("does not dedupe requests when it is disabled", done => {
```
asyncanup: suggestion:
`... when request de-duplication is disabled by the model ...`

test/internal/request/RequestQueue.get.spec.js
```diff
@@ -238,75 +244,175 @@ describe('#get', function() {
...
+            expect(dsGetSpy).toHaveBeenNthCalledWith(2, expect.anything(), [videos1, videos2]);
+
+            done();
+        });
+
+        queue.get([videos0, videos1], [videos0, videos1], zip);
+        queue.get([videos1, videos2], [videos1, videos2], zip);
+    });
+
+    it("combines batched paths without collapse when it is disables", done => {
```
asyncanup: suggestion: `when path collapsing is disabled on the model`

test/internal/request/RequestQueue.get.spec.js
```diff
@@ -238,75 +244,175 @@ describe('#get', function() {
...
+        const scheduler = new ASAPScheduler();
```
asyncanup: what's the relevance, if any, of `ImmediateScheduler` vs `ASAPScheduler`?

asyncanup:
Other than the minor nitpicks I made, looks good!

test/internal/request/RequestQueue.get.spec.js
jcranendonk: - ImmediateScheduler: executes `flushGetRequest` synchronously, never batches (because they'll be marked 'sent' directly after the respective `model.get()` call)
- ASAPScheduler: schedules fetches on the next tick, successive `model.get()` calls are batched

PR #963, opened 2019-07-12
Cleaned up test files and fixtures

jcranendonk:
- Replaced require()-based test loading with [standard Jest config](https://jestjs.io/docs/en/configuration#testmatch-array-string)
- Removed unreferenced test fixtures
- Addressed open comments from https://github.com/Netflix/falcor/pull/962

Same tests, same test coverage, but now it takes 6s instead of 10s (caveat: "on my machine").


asyncanup:
Love this whole PR so much

PR #969, opened 2019-11-12
Remove asap package

jcranendonk:
The asap package (and its fork, falcor-asap) have occasionally caused issues with our build pipeline. We're not using the scheduler that relies on this package, so I'm removing it.

I kept the ASAPScheduler in the repository, in case anyone imports that file, but it simply forwards to `new TimeoutScheduler(1)` now. That scheduler may have slightly different behavior, but it should be similar enough for most use cases.

If you still need the exact `asap` behavior, there is a workaround: you can provide a custom scheduler to Falcor.Model.


lib/schedulers/ASAPScheduler.js
```diff
@@ -1,19 +1,8 @@
...
-ASAPScheduler.prototype.scheduleWithState = function scheduleWithState(state, action) {
-    var self = this;
-    asap(function() {
-        action(self, state);
-    });
-    return empty;
-};
+// Retained for backwards compatibility
+function ASAPScheduler() {
+    return TimeoutScheduler.bind(this, 1)();
```
tquetano-netflix: Can't this just be `TimeoutScheduler.call(this, 1);`?

jcranendonk: ...yes, it can 🤦‍♂

PR #970, opened 2020-01-22
Added request options for DataSource retry info

jcranendonk:
The DataSource for a Model has no insight into the retry mechanism for get and set requests. This PR adds the retry count as one of a set of request options that are passed to `DataSource.get` and `DataSource.set`, so that this information may be used to _e.g._ manage request balancing and load shedding on the underlying network resource.


lib/response/set/setRequestCycle.js
```diff
@@ -17,7 +17,7 @@ module.exports = function setRequestCycle(model, observer, groups,
     var requestedPaths = requestedAndOptimizedPaths.requestedPaths;
 
     // we have exceeded the maximum retry limit.
-    if (count === model._maxRetries) {
+    if (count > model._maxRetries) {
```
tquetano-netflix: Just asking considering the original comparison logic, should this be `>=`?

jcranendonk: Yeah, that'd be safer. I made the change to align the retry counts with `getRequestCycle`; one started at 0, the other at 1. Starting at 1 lines up better with the Zuul retry header value.

lib/response/set/SetResponse.js
```diff
@@ -84,7 +84,7 @@ SetResponse.prototype._subscribe = function _subscribe(observer) {
 
     // Starts the async request cycle.
     return setRequestCycle(
-        model, observer, groups, isJSONGraph, isProgressive, 0);
+        model, observer, groups, isJSONGraph, isProgressive, 1);
```
jcranendonk: @tquetano-netflix The initial value was changed here, in coordination with the retry logic condition.

PR #980, opened 2020-08-03
Minor changes to reduce garbage

bendmorris:
cc: @jcranendonk

The goal here was to reduce the amount being allocated when fetching data using falcor. Our JS engine is particularly sensitive to garbage generation which may not be a problem on modern browsers. There should be no functional changes.

Locally my test case is showing about a 30% reduction in total allocations from Falcor, particularly coming from setJSONGraphPathSet which was reduced by over 50%.

I'll add comments inline to explain the rationale for the changes.


bendmorris:
So that I don't have to call it out everywhere, there were a few utility methods - arrayConcat, fastCat, arraySlice, arrayMap... - which just reimplemented prototype methods with no apparent functional differences. I removed these and replaced them with the native versions; they're not faster, so there's no reason to ship the extra code.

lib/ModelDataSourceAdapter.js
```diff
@@ -11,7 +11,8 @@ ModelDataSourceAdapter.prototype.set = function set(jsongResponse) {
 };
 
 ModelDataSourceAdapter.prototype.call = function call(path, args, suffixes, paths) {
-    var params = [path, args, suffixes].concat(paths);
```
bendmorris, line 14: This creates two arrays - one for the literal, one from concat - when we only need one.

lib/get/onMissing.js
```diff
@@ -33,11 +28,13 @@ module.exports = function onMissing(model, path, depth,
 
 function concatAndInsertMissing(model, remainingPath, depth, requestedPath,
                                 optimizedPath, optimizedLength, results) {
-    results.requestedMissingPaths[results.requestedMissingPaths.length] =
-        fastCat(arraySlice(requestedPath, 0, depth), remainingPath);
```
bendmorris, line 37: This creates a slice just to pass it to concat, resulting in an extra array allocation.

lib/get/util/clone.js
```diff
@@ -6,13 +6,9 @@ module.exports = function clone(node) {
         return node;
     }
 
-    var outValue, i, len;
-    var keys = Object.keys(node);
-    outValue = {};
-    for (i = 0, len = keys.length; i < len; i++) {
-        var k = keys[i];
-        var k0 = k.substr(0, 2);
```
bendmorris, line 14: Both the `Object.keys` (extra array) and the `substr` (extra string) here are unnecessary.

lib/request/complement.js
```diff
@@ -25,9 +23,9 @@ module.exports = function complement(requested, optimized, tree) {
         var subTree = tree[oPath.length];
 
         var intersectionData = findPartialIntersections(rPath, oPath, subTree);
-        intersection = arrayConcat(intersection, intersectionData[0]);
-        optimizedComplement = arrayConcat(optimizedComplement, intersectionData[1]);
-        requestedComplement = arrayConcat(requestedComplement, intersectionData[2]);
```
bendmorris, line 30: Since these arrays are local, and we're reassigning them to the return value of concat, we can just mutate the original arrays instead of creating new ones.

lib/request/complement.js
```diff
@@ -79,96 +77,135 @@ function findPartialIntersections(requestedPath, optimizedPath, requestTree) {
...
     } else if (depthDiff > 0) {
-        return recurse(requestTree, 0, arraySlice(requestedPath, 0, depthDiff), []);
+        return recurse(requestedPath, optimizedPath, requestTree, 0, requestedPath.slice(0, depthDiff), []);
     } else {
-        return recurse(requestTree, -depthDiff, [], arraySlice(optimizedPath, 0, -depthDiff));
+        return recurse(requestedPath, optimizedPath, requestTree, -depthDiff, [], optimizedPath.slice(0, -depthDiff));
     }
+}
 
-    function recurse(currentTree, depth, rCurrentPath, oCurrentPath) {
```
bendmorris, line 89: The change here is that `recurse` becomes a function declaration instead of a closure. It seems to only close over two variables which are now passed in directly. This saves a function object per call of `findPartialIntersections` + an activation per call of `recurse`.

lib/request/complement.js
```diff
@@ -79,96 +77,135 @@ function findPartialIntersections(requestedPath, optimizedPath, requestTree) {
...
+                    oComplementPaths[oComplementPaths.length] = arrayConcatSlice2(
                         oCurrentPath,
-                        arraySlice(optimizedPath, depth + 1)
+                        innerKey,
+                        optimizedPath, depth + 1
                     );
-                    rComplementPaths[rComplementPaths.length] = arrayConcat(
+                    rComplementPaths[rComplementPaths.length] = arrayConcatSlice2(
                         rCurrentPath,
-                        arraySlice(requestedPath, depth + depthDiff + 1)
```
bendmorris, line 157: More `concat` of `slice`. I added a utility below to avoid the extra arrays in each of these cases.

lib/set/setJSONGraphs.js
```diff
@@ -118,18 +118,23 @@ function setJSONGraphPathSet(
 }
 /* eslint-enable */
 
+var _result = new Array(4);
 function setReference(
     root, node, messageRoot, message, requestedPath, optimizedPath,
     version, expired, lru, comparator, errorSelector, replacedPaths) {
 
     var reference = node.value;
-    optimizedPath.splice(0, optimizedPath.length);
```
bendmorris, line 126: `splice` returns an array. Our engine isn't smart enough to optimize it out based on whether you're using it. `length = 0` is better for clearing the array.

lib/set/setJSONGraphs.js
```diff
@@ -118,18 +118,23 @@ function setJSONGraphPathSet(
 }
 /* eslint-enable */
 
+var _result = new Array(4);
```
bendmorris, line 121: This one is admittedly a little ugly. I noticed that the return values of these local functions that return arrays of 4 elements never live long enough to overlap with another call. This means we can just reuse a single array for their return value instead of creating a new one each time.

lib/support/insertNode.js
```diff
@@ -1,17 +1,19 @@
...
 
     if (version !== undefined) {
-        // eslint-disable-next-line camelcase
         node.$_version = version;
     }
-    // eslint-disable-next-line camelcase
     if (!node.$_absolutePath) {
-        // eslint-disable-next-line camelcase
-        node.$_absolutePath = optimizedPath.slice(0, optimizedPath.index).concat(key);
+        if (Array.isArray(key)) {
```
bendmorris, line 10: The original code was apparently calling `concat` with a single element as the argument, TIL you can do that.

I'm not sure if `key` is ever an array, added this block for parity just in case.

PR #989, opened 2021-03-27
Handling fromPath errors in SetResponse constructor

chourihan:
When creating a new `SetResponse` instance, the constructor may make calls to `pathSyntax.fromPath`. If the presumed path is invalid, [the parser may throw an exception](https://github.com/Netflix/falcor-path-syntax/blob/master/src/parse-tree/head.js#L9) which since done in the constructor is uncaught. This catches those exception and emits an error on `_subscribe`.


jcranendonk:
Good catch, thanks for the PR. Could you add a unit test for this scenario? Does this issue exist in GetResponse as well?

jcranendonk:
On looking into SetResponse, it appears that input validation should be handled in lib/support/validateInput.js. Can you verify whether the faulty input is handled correctly in that function, and if not, fix it in that location?

chourihan:
> On looking into SetResponse, it appears that input validation should be handled in lib/support/validateInput.js. Can you verify whether the faulty input is handled correctly in that function, and if not, fix it in that location?

@jcranendonk it does look like this is invoked before the `SetResponse` constructor and is the more appropriate place to catch the error.

chourihan:
@jcranendonk moved the error handling to `lib/support/validateInput.js` and added a couple of tests.

test/Model.spec.js
```diff
@@ -496,4 +496,30 @@ describe("Model", () => {
             });
         });
     });
+
+    describe("set and get", () => {
+        it("should throw an error on invalid input when calling Model:get", (done) => {
+            const model = new Model();
+            model.get("{\"foobar\":[]}").subscribe(() => {
+                done(new Error("Did not receive an error when one was expected"));
```
jcranendonk, line 504: This should be moved to the onComplete handler, rather than onNext. (With rx Observable, onComplete and onError are mutually exclusive: either terminates the sequence.)

test/Model.spec.js
```diff
@@ -496,4 +496,30 @@ describe("Model", () => {
...
+            }, (e) => {
+                console.log(e);
+                expect(e instanceof Error).toBe(true);
+                expect(e.message.indexOf("Path syntax validation error")).toBeGreaterThan(-1);
```
jcranendonk, line 508: You could use an inline snapshot here, it would provide better feedback in case of failure:
https://jestjs.io/docs/expect#tomatchinlinesnapshotpropertymatchers-inlinesnapshot
https://jestjs.io/docs/snapshot-testing#inline-snapshots

jcranendonk (approved):

